PR #3, opened 2018-06-02
Change default fill_mode from nearest to constant

ozabluda:
The default `fill_mode='nearest'` (which was there from the very beginning) for points outside the boundaries of the input is contrary to typical value used (`'constant'`) as well as defaults in SciPy, OpenCV, and everything else I know.

Note that this is different from interpolation for for example for `load_img` (for example see https://github.com/keras-team/keras/pull/8435/)

Also see https://github.com/keras-team/keras/pull/2446

Redirected from PR https://github.com/keras-team/keras/pull/8848

### DIFF
--- a/keras_preprocessing/image.py
+++ b/keras_preprocessing/image.py
@@ -43,7 +43,7 @@ if pil_image is not None:
 
 
 def random_rotation(x, rg, row_axis=1, col_axis=2, channel_axis=0,
-                    fill_mode='nearest', cval=0.):
+                    fill_mode='constant', cval=0.):
     """Performs a random rotation of a Numpy image tensor.
 
     # Arguments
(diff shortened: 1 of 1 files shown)


ozabluda:
@fchollet, @Dref360, Maybe this should be merged now. All TF analogs pad with zeros.

https://www.tensorflow.org/api_docs/python/tf/contrib/image/rotate
https://www.tensorflow.org/versions/r2.0/api_docs/python/tf/image/resize_with_pad
https://www.tensorflow.org/versions/r2.0/api_docs/python/tf/image/resize_with_crop_or_pad
https://github.com/tensorflow/addons/blob/master/tensorflow_addons/image/transform_ops.py#L282

Dref360:
Hi!
I have a list of issues similar to this one. All should be resolved with the new API where we will make a major revision. For discussion please see the keras-team/governance repo.

Get Outlook for Android<https://aka.ms/ghei36>

________________________________

fchollet:

Issue #1, opened 2018-05-31
Replacing `if variable not in dict` construction


lzfelix:
Hello, firstly I would like to thank your for this new library which removes the burden from writing some repetitive code, specially regarding text, and let us focus on solving problems instead.

While reading `text.py`, I spotted the following construction in two different places:
```python
if variable not in some_dict:
    some_dict[variable] = 1
else:
    some_dict[variable] += 1
```

If `some_dict = defaultdict(int)`, then this code could be replaced by the one-liner `some_dict[variable] += 1`. Why not use it? According to the tests below it is even faster:

```python
In [1]: from collections import defaultdict;
In [2]: simple_dict = dict()
In [3]: def fun(): z = defaultdict(int); z['shoe'] += 1;

# >> Inserting a new element
In [4]: %timeit if 'shoe' not in simple_dict: simple_dict['shoe'] = 1
# 56.6 ns ± 0.0336 ns per loop (mean ± std. dev. of 7 runs, 10000000 loops each)

In [5]: %timeit fun
# 36.1 ns ± 0.00178 ns per loop (mean ± std. dev. of 7 runs, 10000000 loops each)

# >> Fetching existing element
In [6]: %timeit simple_dict['shoe']
# 54.3 ns ± 0.012 ns per loop (mean ± std. dev. of 7 runs, 10000000 loops each)

In [7] ddict = defaultdict(int); ddict['shoe'] += 1
In [8]: %timeit ddict['shoe']
# 58.3 ns ± 0.0122 ns per loop (mean ± std. dev. of 7 runs, 10000000 loops each)

# >> Updating existing element:
In [9]: def fun_2():
     ...:     if 'shoe' not in simple_dict: simple_dict['shoe'] = 1
     ...:     else: simple_dict['shoe'] += 1
     ...:   
In [10]: %timeit fun_2()
# 290 ns ± 0.171 ns per loop (mean ± std. dev. of 7 runs, 1000000 loops each)

In [11]: z = defaultdict(int)
In [12]: %timeit z['shoe'] += 1
# 122 ns ± 0.0166 ns per loop (mean ± std. dev. of 7 runs, 10000000 loops each)
```

lunardog:
Huh... is it a documented feature of Python?

Dref360:
Hi,
This seems reasonable, defaultdict were added in Python 2.5 so we should be fine.
Feel free to submit a PR

PR #4, opened 2018-06-03
Replacing dict by defaultdict

lzfelix:
Replacing the following construction

```python
if variable not in some_dict:
    some_dict[variable] = 1
else:
    some_dict[variable] += 1
```

by `defaultdict` instead: `some_dict[value] += 1`, as discussed in #1.

### DIFF
--- a/keras_preprocessing/text.py
+++ b/keras_preprocessing/text.py
@@ -8,7 +8,7 @@ from __future__ import print_function
 import string
 import sys
 import warnings
-from collections import OrderedDict
+from collections import OrderedDict, defaultdict
 from hashlib import md5
 
 import numpy as np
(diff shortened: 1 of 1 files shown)


keras_preprocessing/text.py
```diff
@@ -8,7 +8,7 @@
 import string
 import sys
 import warnings
-from collections import OrderedDict
+from collections import OrderedDict, defaultdict
```
Dref360: Style : One import per line

lzfelix:
Hi @Dref360 I fixed both issues pointed by you. I hope that everything is okay now, since this is my first PR.

lzfelix:
Great!

PR #9, opened 2018-06-05
Make `lower=True` in Tokenizer affect also the character level tokenization

rakeshchada:
Currently if one sets `lower=True` in the Tokenizer - it comes into effect only for word level tokenization (i.e., `char_level = False`).
The text remains in the original case if `char_level` is set to True.

This PR makes things consistent by converting the text to lower case also when `char_level` is set to `True`.

### DIFF
--- a/keras_preprocessing/text.py
+++ b/keras_preprocessing/text.py
@@ -205,6 +205,8 @@ class Tokenizer(object):
         for text in texts:
             self.document_count += 1
             if self.char_level or isinstance(text, list):
+                if self.lower:
+                    text = text.lower()
                 seq = text
             else:
                 seq = text_to_word_sequence(text,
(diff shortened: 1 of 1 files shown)


Dref360:
* Could this break previous preprocessing or it's a bugfix?
* Could you add some tests?

rakeshchada:
@Dref360 

- IMO, This is a bug fix. It seems like a natural expectation for the text to be converted to lower case when `lower` is set to `True` independent of whether `char_level` is set to `True` or `False`.
- Good call on the tests. It seems like `lower` flag on the tokenizer doesn't have any tests. So I also added those along with the tests that are specific for `char_level` with `lower=True`. Thanks!

A side note:
It seems like things behave a bit weird (independent of this PR) when `char_level` is set to `True` and when a list of list of words is passed as `texts` argument for the `Tokenizer`. It still tokenizes at the word level if a list of list of words is passed even though `char_level` is set to `True`. I didn't want to contaminate this PR by fixing that but if you think that's a useful fix - i can make that in a separate PR.

PR #15, opened 2018-06-15
Added support for RGBA images in keras image preprocessing

justinessert:
Added support for loading and saving rgba images with PIL. 
RGBA images should be handled by all of the transform functions identically to RGB images so no changes were made to the actual transform functions

### DIFF
--- a/keras_preprocessing/image.py
+++ b/keras_preprocessing/image.py
@@ -1587,7 +1587,8 @@ class DirectoryIterator(Iterator):
         image_data_generator: Instance of `ImageDataGenerator`
             to use for random transformations and normalization.
         target_size: tuple of integers, dimensions to resize input images to.
-        color_mode: One of `"rgb"`, `"rgba"`, `"grayscale"`. Color mode to read images.
+        color_mode: One of `"rgb"`, `"rgba"`, `"grayscale"`. 
+            Color mode to read images.
         classes: Optional list of strings, names of subdirectories
             containing images from each class (e.g. `["dogs", "cats"]`).
             It will be computed automatically if not set.
--- a/keras_preprocessing/image.py
+++ b/keras_preprocessing/image.py
@@ -444,16 +444,21 @@ def save_img(path,
         **kwargs: Additional keyword arguments passed to `PIL.Image.save()`.
     """
     img = array_to_img(x, data_format=data_format, scale=scale)
+    if img.mode == 'RGBA' and (file_format == 'jpg' or file_format == 'jpeg'):
+        warnings.warn('The JPG format does not support '
+                      'RGBA images, converting to RGB.')
+        img = img.convert('RGB')
     img.save(path, format=file_format, **kwargs)
 
 
-def load_img(path, grayscale=False, target_size=None,
+def load_img(path, color_mode='rgb', target_size=None,
              interpolation='nearest'):
     """Loads an image into PIL format.
 
     # Arguments
         path: Path to image file.
-        grayscale: Boolean, whether to load the image as grayscale.
+        color_mode: One of "grayscale", "rbg", "rgba". Default: "rgb".
+            The desired image format.
         target_size: Either `None` (default to original size)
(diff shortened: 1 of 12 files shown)


keras_preprocessing/image.py
```diff
@@ -860,9 +864,9 @@ def flow_from_directory(self, directory,
             target_size: Tuple of integers `(height, width)`,
                 default: `(256, 256)`.
                 The dimensions to which all images found will be resized.
-            color_mode: One of "grayscale", "rbg". Default: "rgb".
+            color_mode: One of "grayscale", "rbg", "rgba". Default: "rgb".
                 Whether the images will be converted to
-                have 1 or 3 color channels.
+                have 1, 3, or 4 channels.
```
Dref360: typo

justinessert:
I added RGBA testing everywhere that both RGB and grayscale were tested.
Please rerun the test, the failed test timed out. I'm not sure why that is happening, but everything passes on my end.

justinessert:
Also, I apologize for the many commits, I was not able to run the tests locally due to the limitations of my workspace.

tests/image_test.py
```diff
@@ -30,12 +32,18 @@ def setup_class(cls):
             im = Image.fromarray(imarray.astype('uint8')).convert('RGB')
             rgb_images.append(im)
 
+            bias = np.random.rand(cls.img_w, cls.img_h, 1) * 64
+            variance = np.random.rand(cls.img_w, cls.img_h, 1) * (255 - 64)
```
Dref360: you can reuse the variables above.

tests/image_test.py
```diff
@@ -1,5 +1,6 @@
 import pytest
 from PIL import Image
+from math import ceil
```
Dref360: use np.ceil instead, avoid an import

Dref360:
Some small comments and then we're good

keras_preprocessing/image.py
```diff
@@ -441,16 +444,21 @@ def save_img(path,
...
     img = array_to_img(x, data_format=data_format, scale=scale)
+    if img.mode == 'RGBA' and (file_format == 'jpg' or file_format == 'jpeg'):
+        warnings.warn('The JPG format does not support '
+                      'RGBA images, converting to RGB.')
+        img = img.convert('RGB')
     img.save(path, format=file_format, **kwargs)
 
 
-def load_img(path, grayscale=False, target_size=None,
+def load_img(path, color_mode='rgb', target_size=None,
```
Dref360: Didn't see that the first time sorry. Butm this is a breaking change, please keep grayscale and deprecate it.

keras_preprocessing/image.py
```diff
@@ -471,9 +479,12 @@ def load_img(path, grayscale=False, target_size=None,
...
                           'The use of `array_to_img` requires PIL.')
     img = pil_image.open(path)
-    if grayscale:
+    if color_mode == 'grayscale':
         if img.mode != 'L':
             img = img.convert('L')
+    elif color_mode == 'rgba':
+        if img.mode != 'RGBA':
+            img = img.convert('RGBA')
     else:
```
Dref360: `elif color_mode == 'rgb'` and add a else which throws a ValueError

keras_preprocessing/image.py
```diff
@@ -475,6 +475,10 @@ def load_img(path, color_mode='rgb', target_size=None,
         ImportError: if PIL is not available.
         ValueError: if interpolation method is not supported.
     """
+    if grayscale is True:
+        warnings.warn('grayscale is deprecated. Please use '
+                      'color_mode = "grayscale"')
+        load_img = 'grayscale'
```
justinessert: I'm not sure if this is how you wanted me to handle this? If not, feel free to let me know!

Dref360: Shouldn't it be color_mode='grayscale'?

justinessert: It absolutely should, thanks for pointing that out

PR #18, opened 2018-06-22
Different channel shift values at different image channels

zhenglilei:
The channel shift values for different image channels can be different. So it is incorrect to do a one-time randomization before calling the "apply_channel_shift" function. The changes are made according to the codes in the previous version, Keras 2.1.6, i.e., randomization is done for each channel of an image.

`channel_images = [
        np.clip(x_channel + np.random.uniform(-intensity, intensity),
                min_x,
                max_x)
        for x_channel in x]`

A reference for using channel shift in data augmentation: https://github.com/JustinhoCHN/keras-image-data-augmentation#8-channel_shift_range


Dref360:
The point of `apply_channel_shift` is to be deterministic.

PR #19, opened 2018-06-28
Adapting image.py to 16 bits gray

GuillaumeErhard:
Avoid needless downsampling of 16 bits gray image to 8 bits, and current clipping from Pillow is wrong, see :
https://github.com/python-pillow/Pillow/issues/3011

### DIFF
--- a/keras_preprocessing/image.py
+++ b/keras_preprocessing/image.py
@@ -472,7 +472,7 @@ def load_img(path, grayscale=False, target_size=None,
                           'The use of `array_to_img` requires PIL.')
     img = pil_image.open(path)
     if grayscale:
-        if img.mode != 'L':
+        if img.mode not in ('L', 'I;16'):
             img = img.convert('L')
     else:
         if img.mode != 'RGB':


Dref360:
Please fix conflicts and add a test.

matthewbundala:
@GuillaumeErhard , are you still working on this? I need to load 16-bit single channel tiff images, and ran into the exact issue you are trying to fix with image clipping.

GuillaumeErhard:
@matthewbundala I am lacking the time right now to go further with these changes as right now it is affecting the process for 8 bits image. For example if you use the data augmentation of changing brightness the image is in a float32 bits state and it needs to be casted back to 8 / 16 bits, and it is impossible to access what you started you started with.

On top of that Pillow cannot resize an image with only one mode, nearest so all these limitations should be added in the documentation, but I don't know where and what are all the limitations of Pillow on 16 bits image.

What I think should be done is restarting from scratch in keras-contrib to reimplement all these functions using opencv for unconventional format like gray 16 bits, rgb 16 bits and so on, and keep it clean here with Pillow.

If you want to make it work never the less but loss some functionality, just change in your keras-preprocessing package the file image.py at the function load_img and change the check for gray image from img.mode == 'L' to image.mode in ('L', 'I;16')

Issue #23, opened 2018-07-16
pass `order` to affine_transform


OmriTreidel:
The method `apply_affine_transform` is a wrapper `scipy.ndimage.interpolation.affine_transform` that has a parameter `order` which is the order a a spline interpolation. Not being able to pass this parameter causes a problem when generating random transformation of labeled images since it results in non-integer values.

for example: 

for a given labels (each integer is a class label) 
```python 
[[2 2 0 2 2]
 [1 3 2 3 1]
 [2 1 0 1 2]
 [3 1 0 2 0]
 [3 1 3 2 1]]
```

it transformed to 
```python 
[[2.289865   1.7110896  1.8507836  2.172145   0.15832195]
 [3.         2.2037435  1.0774351  1.4194988  2.3764393 ]
 [3.         1.4194988  0.39426237 0.49894607 1.7452569 ]
 [1.8929222  1.7380519  1.7849773  2.         1.2062019 ]
 [1.2646285  2.8956027  2.2367117  1.3144331  0.23671168]]
```

Which are clearly not valid class labels. I dag around and found that the reason is the hard coded `order` in https://github.com/keras-team/keras-preprocessing/blob/master/keras_preprocessing/image.py line 323. 

being able to pass this parameter would enable correct data augmentation for image segmentation. 

I have made a small path to fix this issue so i'll create a PR soon.

rragundez:
@Dref360 this issue cna be closed as the feature has been merged into master.

PR #24, opened 2018-07-16
interpolation `order` for ImageDataGenerator

OmriTreidel:
### Summary
ImageDataGenerator can take a keyword argument `interpolation_order` that trickles down to `scipy.ndimage.interpolation.affine_transform` that resides in `apply_affine_transform`. At the moment the order of the spline interpolation is hard coded and set to 1 [line 323](https://github.com/keras-team/keras-preprocessing/blob/master/keras_preprocessing/image.py) This causes images of class labels to transform incorrectly. The suggested fix allows to use neaset neighbours interpolation and preserve the pixel value as integers (otherwise it turns to rational numbers that have no meaning as class labels).

for example:

```python
[[2 2 0 2 2]
 [1 3 2 3 1]
 [2 1 0 1 2]
 [3 1 0 2 0]
 [3 1 3 2 1]]
```
is transformed to 
```python
[[2.289865   1.7110896  1.8507836  2.172145   0.15832195]
 [3.         2.2037435  1.0774351  1.4194988  2.3764393 ]
 [3.         1.4194988  0.39426237 0.49894607 1.7452569 ]
 [1.8929222  1.7380519  1.7849773  2.         1.2062019 ]
 [1.2646285  2.8956027  2.2367117  1.3144331  0.23671168]]
```

This has great significance when augmenting data for image segmentation  

### Related Issues
#23 
### PR Overview

- [y] This PR requires to update the documentation [y/n] (make sure the docs are up-to-date) - docstrings have been apdated
- [y] This PR is backwards compatible [y/n] - default values were chosen to preserve current behaviour

### DIFF
--- a/keras_preprocessing/image.py
+++ b/keras_preprocessing/image.py
@@ -51,7 +51,7 @@ if pil_image is not None:
 
 
 def random_rotation(x, rg, row_axis=1, col_axis=2, channel_axis=0,
-                    fill_mode='nearest', cval=0.):
+                    fill_mode='nearest', cval=0., interpolation_order=3):
     """Performs a random rotation of a Numpy image tensor.
 
     # Arguments
@@ -65,18 +65,20 @@ def random_rotation(x, rg, row_axis=1, col_axis=2, channel_axis=0,
             (one of `{'constant', 'nearest', 'reflect', 'wrap'}`).
         cval: Value used for points outside the boundaries
             of the input if `mode='constant'`.
-
+        interpolation_order int: order of spline interpolation.
+            see `ndimage.interpolation.affine_transform`
     # Returns
         Rotated Numpy image tensor.
     """
     theta = np.random.uniform(-rg, rg)
     x = apply_affine_transform(x, theta=theta, channel_axis=channel_axis,
-                               fill_mode=fill_mode, cval=cval)
+                               fill_mode=fill_mode, cval=cval,
+                               order=interpolation_order)
     return x
 
 
 def random_shift(x, wrg, hrg, row_axis=1, col_axis=2, channel_axis=0,
-                 fill_mode='nearest', cval=0.):
+                 fill_mode='nearest', cval=0., interpolation_order=3):
     """Performs a random spatial shift of a Numpy image tensor.
 
     # Arguments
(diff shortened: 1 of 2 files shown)


Dref360:
The default value should be 1 to be backward compatible no?

OmriTreidel:
@Dref360 I've made sure that not setting the order will reproduce end existing behaviour. i.e. order = 1 in `apply_affine_transform` which used to be hard coded( when it was calling `ndimage.interpolation.affine_transform` ). 

When methods like `random_zoom` where called directly, they call `apply_affine_transform` without specifying the interpolation order and therefore call `ndimage.interpolation.affine_transform` with it's default value which is 3 https://docs.scipy.org/doc/scipy-0.14.0/reference/generated/scipy.ndimage.interpolation.affine_transform.html.

@yanmu2017 I'm not sure what you're after. Do you want me to change `order` to `interpolation_order` in  `apply_affine_transform` ?

OmriTreidel:
@Dref360, I don't mind changing it but would would make it backwards **incompatible**.  I doubt that this is what you want. Note that the default value in `apply_affine_transform` is indeed 1 which guarantee backwards compatibility. 

The order = 1 was only forced when calling `apply_affine_transform`. When calling  `random_zoom`  you force it to be 3, which is very unexpected. However, It's probably better to not to alter the existing behaviour.

Dref360:
Using your branch we do get a different behavior if I don't set the interpolation order to 1 manually.

```python
import keras
import numpy as np
from keras_preprocessing.image import random_zoom
from scipy.stats import describe

np.random.seed(1337)
x = np.random.normal(size=[224,224,3])
x1 = random_zoom(x, (0.8, 1.2),channel_axis=-1,interpolation_order=1) # Need interpolation_order=1 to reproduce the result.

print(describe(x1.reshape([-1])))
```

## On keras master

```
Using TensorFlow backend.
DescribeResult(nobs=150528, minmax=(-3.5309916438038758, 3.093515040110392), mean=0.0023176740798912885, variance=0.49648973654639206, skewness=-0.02221715888068403, kurtosis=0.31006682949297426)

Process finished with exit code 0
```

## On your branch without interpolation_order=1

```
Using TensorFlow backend.
DescribeResult(nobs=150528, minmax=(-3.826666024899408, 3.7900867414241937), mean=0.0022082395056552575, variance=0.7849180825381086, skewness=-0.015615107020020502, kurtosis=0.038251322426397394)

Process finished with exit code 0
```

## With interpolation_order=1

```
Using TensorFlow backend.
DescribeResult(nobs=150528, minmax=(-3.5309916438038758, 3.093515040110392), mean=0.0023176740798912885, variance=0.49648973654639206, skewness=-0.02221715888068403, kurtosis=0.31006682949297426)

Process finished with exit code 0
```

OmriTreidel:
You're absolutely right. I'll make the change

OmriTreidel:
You're absolutely right. I've made the change.

Dref360:
Please fix PEP8

PR #29, opened 2018-07-27
Serialize text and sequence preprocessing

maxpumperla:
### Summary

Serde for text and sequence preprocessing (will do image in a follow-up PR). Serializes objects to JSON in usual Keras-style, comes with unit tests.

Note to reviewer: Where possible we store data as plain text. This is done so that other projects (not necessarily Python) can pick up the configs and restore them. Particularly, I'll write preprocessor loaders for DL4J in Java.

Autogen docs should pick up these additions, but I'm not 100% how this works with this separate project.
---

- [x] This PR requires has unit tests
- [x] This PR requires no update to documentation
- [x] This PR is backwards compatible
- [x] This PR adds new functionality to the API

### DIFF
--- a/keras_preprocessing/text.py
+++ b/keras_preprocessing/text.py
@@ -458,7 +458,7 @@ class Tokenizer(object):
             'index_docs': json_index_docs
         }
 
-    def to_json(self):
+    def to_json(self, **kwargs):
         """Returns a JSON string containing the tokenizer configuration.
         To load a tokenizer from a JSON string, use
         `keras.preprocessing.text.tokenizer_from_json(json_string)`.
@@ -475,7 +475,7 @@ class Tokenizer(object):
             'class_name': self.__class__.__name__,
             'config': config
         }
-        return json.dumps(tokenizer_config)
+        return json.dumps(tokenizer_config, **kwargs)
 
 
 def tokenizer_from_json(json_string):
--- a/keras_preprocessing/text.py
+++ b/keras_preprocessing/text.py
@@ -501,9 +501,9 @@ def tokenizer_from_json(json_string):
     word_docs = json.loads(config.pop('word_docs'))
     index_docs = json.loads(config.pop('index_docs'))
     # Integer indexing gets converted to strings with json.dumps()
-    index_docs = {int(k):v for k,v in index_docs.items()}
+    index_docs = {int(k): v for k, v in index_docs.items()}
     index_word = json.loads(config.pop('index_word'))
-    index_word = {int(k):v for k,v in index_word.items()}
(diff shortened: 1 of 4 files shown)


maxpumperla:
@fchollet any chance this could get reviewed?

tests/sequence_test.py
```diff
@@ -103,6 +103,18 @@ def test_remove_long_seq():
...
+def test_TimeseriesGenerator_serde():
+    data = np.array([[i] for i in range(50)])
+    targets = np.array([[i] for i in range(50)])
+
+    data_gen = sequence.TimeseriesGenerator(data, targets,
+                                            length=10,
+                                            sampling_rate=2,
+                                            batch_size=2)
+    json_gen = data_gen.to_json()
+    recovered_gen = sequence.timeseries_generator_from_json(json_gen)
```
Dref360: please assert that the properties are the equals.

maxpumperla:
@Dref360 I already did, but maybe I should do so more clearly. Thanks for your review, will refine the test later.

People outside academia need to persist stuff to ship it to production, store it somewhere etc. This  holds true for Keras models (I introduced yaml and json serialization back in 2015 for Keras models), but would be nice for preprocessing and "everything else" you need in your pipeline.

The concrete use case is this: I'm the author of Keras model import for DL4J. We currently have a few requests of users wanting to bring their models there, but lack the preprocessing for various cases. Instead of configuring things twice, I want to offer e2e support for import. (I know I can pickle e.g. a `Tokenizer`, but that doesn't help me in java/scala/whatever).

Regardless of this concrete use case, I still see this as beneficial overall. It's even possible to consider storing preprocessing configs in HDF5 groups in `model.save()` or so.

Dref360:
All good, once the tests are fixed it would be lgtm

maxpumperla:
@Dref360 good to go now

keras_preprocessing/text.py
```diff
@@ -429,3 +433,84 @@ def sequences_to_matrix(self, sequences, mode='binary'):
...
+    config = tokenizer_config.get('config')
+
+    word_counts = json.loads(config.pop('word_counts'))
+    word_docs = json.loads(config.pop('word_docs'))
+    index_docs = json.loads(config.pop('index_docs'))
+    # Integer indexing gets converted to strings with json.dumps()
+    index_docs = {int(k): v for k, v in index_docs.items()}
+    index_word = json.loads(config.pop('index_word'))
+    index_word = {int(k): v for k, v in index_word.items()}
+    word_index = json.loads(config.pop('word_index'))
```
michal-demecki-profil-software: index_docs and word docs should be instance of 'defaultdict'

PR #30, opened 2018-07-28
added tiff as an extension for listing images

elmerehbi:


Dref360:
This looks like deadcode to me? Should we just deprecate it?
What's your usecase to keep it?

We could keep it but we would need to add tests and docs.

elmerehbi:
@Dref360 Why deadcode? No one  is using tiff?

My use case is loading 16bit tiff (grayscale satellite) images.

Is there a better way?

Dref360:
I was talking about the fact that list_pictures is not used anywhere in the codebase.

elmerehbi:
@Dref360 Ah I didn't notice that. I am not using it either (using PIL instead) but thought it would be a good addition. However following your comment then I agree.

PR #32, opened 2018-08-03
Include tests in PyPI tarball

dotlambda:
### Summary
This is useful for downstream distributions to test if the package works correctly with their build recipe. Also, it provides a way to test if the dependencies as they are packaged by the distribution are compatible.

### PR Overview

### DIFF
--- a/MANIFEST.in
+++ b/MANIFEST.in
@@ -1,0 +1,1 @@
+include tests/*.py


Dref360:
We should mimic the MANIFEST of keras
https://github.com/keras-team/keras/blob/master/MANIFEST.in

Dref360:
Done in https://github.com/keras-team/keras-preprocessing/pull/71

PR #36, opened 2018-08-14
adds flow_from_dataframe to ImageDataGenerator

Vijayabhaskar96:
### Summary
This PR adds the flow_from_dataframe method to the ImageDataGenerator class, which allows the users to read the image files directly from the dataframe which maps the filenames to the classe names. Since, the data is read from dataframe directly, the ImageDataGenerator can be used for regression tasks too.

If the target values are present in the dataframe they can be directly read from flow_from_dataframe function by specifing the column (list of column names,If multiple target values) that has the target data.
        Takes the dataframe and the path to a directory
         and generates batches of augmented/normalized data.

        # A simple tutorial can be found at: http://bit.ly/keras_flow_from_dataframe

        # Arguments
                dataframe: Pandas dataframe.
                directory: string,path to the target directory that contains all
                           the images mapped in the dataframe.
                x_col: string,column in the dataframe that contains
                       the filenames of the target images.
                y_col: string or list of strings,columns in
                       the dataframe that will be the target data.
                has_ext: bool, True if filenames in dataframe[x_col]
                        has filename extensions,else False.
                target_size: tuple of integers `(height, width)`,
                             default: `(256, 256)`.
                             The dimensions to which all images
                             found will be resized.
                color_mode: one of "grayscale", "rbg". Default: "rgb".
                            Whether the images will be converted to have
                            1 or 3 color channels.
                classes: optional list of classes
                (e.g. `['dogs', 'cats']`). Default: None.
                 If not provided, the list of classes will be automatically
                 inferred from the y_col,
                 which will map to the label indices, will be alphanumeric).
                 The dictionary containing the mapping from class names to class
                 indices can be obtained via the attribute `class_indices`.
                class_mode: one of "categorical", "binary", "sparse",
                  "input", "other" or None. Default: "categorical".
                 Determines the type of label arrays that are returned:
                 - `"categorical"` will be 2D one-hot encoded labels,
                 - `"binary"` will be 1D binary labels,
                 - `"sparse"` will be 1D integer labels,
                 - `"input"` will be images identical
                 to input images (mainly used to work with autoencoders).
                - `"other"` will be numpy array of y_col data
                 - None, no labels are returned (the generator will only
                         yield batches of image data, which is useful to use
                 `model.predict_generator()`, `model.evaluate_generator()`, etc.).
                batch_size: size of the batches of data (default: 32).
                shuffle: whether to shuffle the data (default: True)
                seed: optional random seed for shuffling and transformations.
                save_to_dir: None or str (default: None).
                             This allows you to optionally specify a directory
                             to which to save the augmented pictures being generated
                             (useful for visualizing what you are doing).
                save_prefix: str. Prefix to use for filenames of saved pictures
                (only relevant if `save_to_dir` is set).
                save_format: one of "png", "jpeg"
                (only relevant if `save_to_dir` is set). Default: "png".
                follow_links: whether to follow symlinks inside class subdirectories
                (default: False).
                subset: Subset of data (`"training"` or `"validation"`) if
                 `validation_split` is set in `ImageDataGenerator`.
                interpolation: Interpolation method used to resample the image if the
                 target size is different from that of the loaded image.
                 Supported methods are `"nearest"`, `"bilinear"`, and `"bicubic"`.
                 If PIL version 1.1.3 or newer is installed, `"lanczos"` is also
                 supported. If PIL version 3.4.0 or newer is installed, `"box"` and
                 `"hamming"` are also supported. By default, `"nearest"` is used.

### Related Issues
Adds functionality referenced here:
https://github.com/keras-team/keras/issues/10069
Followed the feedbacks from PR https://github.com/keras-team/keras/pull/10127 
### PR Overview


keras_preprocessing/image.py
```diff
@@ -670,6 +670,49 @@ class ImageDataGenerator(object):
...
+            x_col='filenames',
+            y_col='classes'
+            has_ext=True,
+            target_size=(150, 150),
+            batch_size=32,
+            class_mode='binary')
+
+    df_valid=read_csv("./test.csv")
+
+    validation_generator = test_datagen.flow_from_directory(
```
Dref360: Should be flow_from_dataframe

keras_preprocessing/image.py
```diff
@@ -967,6 +1010,97 @@ class subdirectories (default: False).
...
+                            save_format='png',
+                            subset=None,
+                            interpolation='nearest'):
+        """Takes the dataframe and the path to a directory
+         and generates batches of augmented/normalized data.
+
+        # A simple tutorial can be found at: http://bit.ly/keras_flow_from_dataframe
+
+        # Arguments
+                dataframe: Pandas dataframe.
```
Dref360: Please specify the format of the Dataframe. It's not clear from the doc.

keras_preprocessing/image.py
```diff
@@ -1656,7 +1803,8 @@ class DirectoryIterator(Iterator):
             "hamming" are also supported. By default, "nearest" is used.
     """
 
-    def __init__(self, directory, image_data_generator,
+    def __init__(self, image_data_generator, dataframe, directory,
```
Dref360: The order doesn't match the one in the doc.

keras_preprocessing/image.py
```diff
@@ -1665,6 +1813,18 @@ def __init__(self, directory, image_data_generator,
                  follow_links=False,
                  subset=None,
                  interpolation='nearest'):
+        if dataframe is not None:
```
Dref360: I would prefer if those 2 cases were in different classes. This is not easily maintainable.

Dref360: Maybe we should do a base class with abstract methods that get the classes, samples and filenames.

Vijayabhaskar96: I'm thinking something like a new ImageFileIterator as a Class inherited from Iterator class, which handles all the common stuff except classes, filenames, sample,class_mode and class_indices, and Finally DirectoryIterator and a DataFrameIterator inheriting from ImageFileIterator. Or we can make the Iterator class itself handle all the common stuff, and DirectoryIterator and a DataFrameIterator inheriting from it.

Dref360: Whichever methods you think is the best. I just don't think we should have any :+1: 

```
if dataframe is not None:
   # pandas stuff
elif h5pystuff is not None:
   # H5PY stuff
else:
   # Flow from directory
```

I think it would be better to have one class per Iterator type that handle all everything specific with this Iterator.

Vijayabhaskar96: I have updated my PR, Please have a look.

keras_preprocessing/image.py
```diff
@@ -1820,3 +1936,255 @@ def next(self):
...
+            print('Found %d images belonging to %d classes.' %
+                  (self.samples, self.num_classes))
+        else:
+            print('Found %d images.' % self.samples)
+        # Second, build an index of the images
+        # in the different class subfolders.
+        results = []
+        self.filenames = []
+        self.classes = np.zeros((self.samples,), dtype='int32')
+        if dataframe is None:
```
Dref360: When is dataframe equals to None?

SWeichenthal:
Is there any way to make this available for keras in R? I apologize if this is a ridiculous question but this would be very helpful to have.

keras_preprocessing/image.py
```diff
@@ -1820,3 +1936,233 @@ def next(self):
...
+        if self.num_classes > 0:
+            print('Found %d images belonging to %d classes.' %
+                  (self.samples, self.num_classes))
+        else:
+            print('Found %d images.' % self.samples)
+
+        # Second, build an index of the images.
+        self.filenames = []
+        self.classes = np.zeros((self.samples,), dtype='int32')
+        filenames = _list_valid_filenames_in_directory(
```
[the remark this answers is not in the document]
Vijayabhaskar96: All the valid images are found first and then the images specified in the dataframe are only used later, it should work when dataframe has less no. of images than in directory. If you can reproduce the the problem, create an issue,I will look into it.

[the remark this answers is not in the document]
apollo2030: Sure, [here is the link to the issue](https://github.com/keras-team/keras-preprocessing/issues/60). 

Besides, thanks a lot for this work!

PR #37, opened 2018-08-14
pad_sequences: Add support for string value.

juharris:
### Summary
Add support for `value` parameter as a string to `pad_sequences`.

### PR Overview

### DIFF
--- a/keras_preprocessing/sequence.py
+++ b/keras_preprocessing/sequence.py
@@ -9,6 +9,8 @@ import numpy as np
 import random
 import json
 from six.moves import range
+import six
+import warnings
 
 from . import get_keras_submodule
 
@@ -74,6 +76,13 @@ def pad_sequences(sequences, maxlen=None, dtype='int32',
         if len(s) > 0:
             sample_shape = np.asarray(s).shape[1:]
             break
+
+    if isinstance(value, six.string_types) and dtype != object \
+            and not np.issubdtype(dtype, np.str_) \
+            and not np.issubdtype(dtype, np.unicode_):
+        warnings.warn("`dtype` {} is not compatible with `value`'s type: {}\n"
+                      "You should set `dtype=object` for variable length strings."
+                      .format(dtype, type(value)))
 
     x = np.full((num_samples, maxlen) + sample_shape, value, dtype=dtype)
     for idx, s in enumerate(sequences):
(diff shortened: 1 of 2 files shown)


keras_preprocessing/sequence.py
```diff
@@ -39,12 +39,13 @@ def pad_sequences(sequences, maxlen=None, dtype='int32',
...
         maxlen: Int, maximum length of all sequences.
         dtype: Type of the output sequences.
+            To pad sequences with variable length strings, you can use `object`.
         padding: String, 'pre' or 'post':
             pad either before or after each sequence.
         truncating: String, 'pre' or 'post':
             remove values from sequences larger than
             `maxlen`, either at the beginning or at the end of the sequences.
-        value: Float, padding value.
+        value: Float or String, padding value.
```
Dref360: We would need some error handling.
If the users set value='pad', but doesn't set `dtype`, they would get a cryptic error message.

juharris: Thanks. Good point.
It can be tricky to check this and handle all of the cases of incompatible dtypes. Do you have any suggestions on a nice way to check for errors?

juharris: I added a warning.

keras_preprocessing/sequence.py
```diff
@@ -74,7 +77,14 @@ def pad_sequences(sequences, maxlen=None, dtype='int32',
             sample_shape = np.asarray(s).shape[1:]
             break
 
-    x = (np.ones((num_samples, maxlen) + sample_shape) * value).astype(dtype)
+    if isinstance(value, six.string_types) and dtype != object \
+            and not np.issubdtype(dtype, np.str_) \
+            and not np.issubdtype(dtype, np.unicode_):
+        warnings.warn("`dtype` {} is not compatible with `value`'s type: {}\n"
```
Dref360: Shouldn't we throw?
Also, we try to avoid \  when possible. You can create a variable `is_supported = dtype != object` and add more conditions.

juharris: Good points.
I didn't want to throw and block people that might be using another dtype that is okay even if it fails my check because of something I didn't realize. I'm happy to change it to throw if you'd prefer it.

tests/sequence_test.py
```diff
@@ -35,6 +36,33 @@ def test_pad_sequences():
...
+
+    # test truncating
+    b = sequence.pad_sequences(a, maxlen=2, truncating='pre', value='pad',
+                               dtype=object)
+    assert_equal(b, [['pad', '1'], ['1', '2'], ['2', '3']])
+    b = sequence.pad_sequences(a, maxlen=2, truncating='post', value='pad',
+                               dtype='<U3')
+    assert_equal(b, [['pad', '1'], ['1', '2'], ['1', '2']])
+
+    try:
```
Dref360: final fix : 
Use with pytest.raises(ValueError) as err:

Dref360:
LGTM thx

PR #46, opened 2018-08-31
Ensure that ImageDataGenerator.flow() generates stratified training a…

mfridental:
…nd test sets even if the passed numpy arrays are sorted by class

### Summary
A small improvement that fixes a problem that could take days to debug. 

In case the initial numpy arrays used for ImageDataGenerator.flow() are sorted by class, which can be the case if the user's code is reading images class by class, there is a high chance that the class frequencies in the resulting training and tests will differ. In the worst case, the training set will contain only samples of one class and the test set samples of another class, which would lead to poor model generalization.

The proposed patch fixes the problem by shuffling numpy arrays before they are split into training and validation.

### Related Issues
The issue happened to me at work. I've fixed it rather than opening an issue.

### PR Overview

- [n] This PR requires new unit tests [y/n] (make sure tests are included) (OLD TEST EXTENDED TO COVER THE FIX)

### DIFF
--- a/keras_preprocessing/image.py
+++ b/keras_preprocessing/image.py
@@ -1554,6 +1554,13 @@ class NumpyArrayIterator(Iterator):
                 raise ValueError('Invalid subset name:', subset,
                                  '; expected "training" or "validation".')
             split_idx = int(len(x) * image_data_generator._validation_split)
+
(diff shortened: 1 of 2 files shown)


mfridental:
Will fix it and make another pull request

PR #48, opened 2018-09-03
Ensure that ImageDataGenerator.flow() generates stratified training and test sets even if the passed numpy arrays are sorted by class

mfridental:
Summary
A small improvement that fixes a problem that could take days to debug.

In case the initial numpy arrays used for ImageDataGenerator.flow() are sorted by class, which can be the case if the user's code is reading images class by class, there is a high chance that the class frequencies in the resulting training and tests will differ. In the worst case, the training set will contain only samples of one class and the test set samples of another class, which would lead to poor model generalization.

The proposed patch fixes the problem by shuffling numpy arrays before they are split into training and validation.

Related Issues
The issue happened to me at work. I've fixed it rather than opening an issue.

PR Overview
•[n] This PR requires new unit tests [y/n] (make sure tests are included) (OLD TEST EXTENDED TO COVER THE FIX)
•[n] This PR requires to update the documentation [y/n] (make sure the docs are up-to-date)
•[y] This PR is backwards compatible [y/n]
•[n] This PR changes the current API [y/n] (all API changes need to be approved by fchollet)

### DIFF
--- a/keras_preprocessing/image.py
+++ b/keras_preprocessing/image.py
@@ -1555,13 +1555,13 @@ class NumpyArrayIterator(Iterator):
                                  '; expected "training" or "validation".')
             split_idx = int(len(x) * image_data_generator._validation_split)
 
-            # in case the numpy arrays are sorted by class, we need to shuffle them,
-            # to ensure that both training and test set will get a stratified sample
-            if seed is not None:
-                np.random.seed(seed)
-            permutation_idx = np.random.permutation(len(x))
-            x = x[permutation_idx]
-            y = y[permutation_idx]
+            if len(np.unique(y)) != len(np.unique(y[:split_idx])) or \
+                    len(np.unique(y)) != len(np.unique(y[split_idx:])):
+                raise ValueError('Training and validation subsets would '
+                                 'have different number of classes after '
+                                 'the split. If your numpy arrays are '
+                                 'sorted by the label, you might want '
+                                 'to shuffle them.')
 
             if subset == 'validation':
                 x = x[:split_idx]
@@ -1573,6 +1573,7 @@ class NumpyArrayIterator(Iterator):
                 x_misc = [np.asarray(xx[split_idx:]) for xx in x_misc]
                 if y is not None:
                     y = y[split_idx:]
+
         self.x = np.asarray(x, dtype=self.dtype)
         self.x_misc = x_misc
         if self.x.ndim != 4:
--- a/keras_preprocessing/image.py
+++ b/keras_preprocessing/image.py
@@ -1557,6 +1557,8 @@ class NumpyArrayIterator(Iterator):
 
             # in case the numpy arrays are sorted by class, we need to shuffle them,
             # to ensure that both training and test set will get a stratified sample
+            if seed is not None:
+                np.random.seed(seed)
             permutation_idx = np.random.permutation(len(x))
             x = x[permutation_idx]
             y = y[permutation_idx]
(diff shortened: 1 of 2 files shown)


keras_preprocessing/image.py
```diff
@@ -1554,6 +1554,13 @@ def __init__(self, x, y, image_data_generator,
                 raise ValueError('Invalid subset name:', subset,
                                  '; expected "training" or "validation".')
             split_idx = int(len(x) * image_data_generator._validation_split)
+
+            # in case the numpy arrays are sorted by class, we need to shuffle them,
```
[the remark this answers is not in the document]
mfridental: You're right, adding seed parameter support

Dref360:
If we shuffle every time, wouldn't this have for effect that some samples will be in the training set and the validation set?

mfridental:
Hmm. 

Either we educate users to pass the same seed for training and validation sets, in this case these sets will be disjunct. 

Or we store the permutation_idx inside of the generator - in that case I would move the generation of subsets from NumpyArrayIterator to the generator (in case NumpyArrayIterator is not considered to be part of the public API). This would be not consistent with other two iterators though. 

What would you prefer?

Dref360:
I would prefer to just assert that np.unique(y) == np.unique(y_val) == np.unique(y_train) and raises an error if it's not the case. This would achieve the same thing as your idea.

mfridental:
Uhm, I'm not entirely clear how this assert can ensure that no duplicated Xes will appear in the train and validation sets after the permutation. 

Or you think we shouldn't permute at all, but only warn the user? In that case we would need to check the class frequences, no? Because if just one sample of some class would be present in the train set, it would satisfy the assertion, but still lead to very poor model generalization.

mfridental:
So how do we proceed from here?

Dref360:
I would just raise an error (or a warning).
The issue we want to fix is that we want every class in every set.

I'm not sure if the proportions matter here. We could consider it if we want to raise a warning, but it could become chaotic if we raise an error.

keras_preprocessing/image.py
```diff
@@ -1554,6 +1554,15 @@ def __init__(self, x, y, image_data_generator,
                 raise ValueError('Invalid subset name:', subset,
                                  '; expected "training" or "validation".')
             split_idx = int(len(x) * image_data_generator._validation_split)
+
+            if len(np.unique(y)) != len(np.unique(y[:split_idx])) or \
```
Dref360: By construction, if np.unique(y) == np.unique(y_val) and np.unique(y) == np.unique(y_train) then np.unique(y_val) == np.unique(y_train). So we only need to check that np.unique(y_train) == np.unique(y_val).

Also to check for equality you can use `np.array_equal`.
`np.array_equal(np.unique(x),np.unique(y))`

In any case, we try to avoid \.

tests/image_test.py
```diff
@@ -200,17 +200,37 @@ def test_image_data_generator_with_validation_split(self):
...
             images = np.vstack(img_list)
+            labels = np.concatenate([
+                np.zeros((int(len(images) / 2),)),
+                np.ones((int(len(images) / 2),))])
             generator = image.ImageDataGenerator(validation_split=0.5)
-            seq = generator.flow(images, np.arange(images.shape[0]),
-                                 shuffle=False, batch_size=3,
+
+            # training and validation sets would have different
+            # number of classes, because labels are sorted
```
Dref360: please check that the error message is the right one.

PR #64, opened 2018-10-06
adding gaussian noise

woolfel:
I wanted to add noise to the training images, but keras didn't have it,
so I decided to implement it

### PR Overview

### DIFF
--- a/keras_preprocessing/image.py
+++ b/keras_preprocessing/image.py
@@ -349,6 +349,34 @@ def flip_axis(x, axis):
     return x
 
 
+def add_gaussian_noise(x, noise_sigma):
+    """ Converts a 3D Numpy array to an image, applies noise and returns tensor.
+    
+    # Arguments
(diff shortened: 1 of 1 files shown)


woolfel:
The image_test.py file is quite big. I'm not sure where to add the test case. I tested it manually. attached is example of the original image and noised image.
noised
image00

Dref360:
In the literature, adding Gaussian noise is not widely used, so I don't see the use case for this feature.

woolfel:
Ok. I can understand why gaussian isn't widely used for things like MNIST or Fashion MNIST. For training faster_rcnn_resnet101 I manually added gaussian noise in Photoshop, which is a tedious manual process. It did reduce overfitting on RGB images larger than 160x120. Thanks for reviewing it, I find it useful so I'll keep in my fork.

fchollet:

PR #61, opened 2018-09-25
fix flow_from_dataframe  no. of files found mismatch bug

Vijayabhaskar96:
### Summary
This PR fixes the bug c and also returns filenames with the extensions for all the class_mode for flow_from_dataframe.
### Related Issues
https://github.com/keras-team/keras-preprocessing/issues/60
### PR Overview

- [ ] This PR requires new unit tests [y/n] (make sure tests are included)
- [ ] This PR requires to update the documentation [y/n] (make sure the docs are up-to-date)
- [x] This PR is backwards compatible [y/n]
- [ ] This PR changes the current API [y/n] (all API changes need to be approved by fchollet)

Issue #60, opened 2018-09-24
flow_from_dataframe generates all the images from directory regardless of x_col


apollo2030:
When the image directory has more files than specified in the 'x_col' of the dataframe, the generator generates more images than expected. See the [repro](https://github.com/apollo2030/Repro/blob/master/Repro.ipynb).

It might be that I don't understand how it works though :)

PR #72, opened 2018-10-17
Fix "DataFrameIterator is regardless of x_col" bug

smurak:
### Summary

- "DataFrameIterator is regardless of x_col" issue (#60) still apears when the input dataframe has only one column(filename column).
  This PR fixes this issue and adds a test.
- This PR also fixes a bug in test_dataframe_iterator_class_mode_input.
  (The current API passes this test because of the previous bug.)

### Related Issues

#60, #61

### PR Overview

### DIFF
--- a/keras_preprocessing/image.py
+++ b/keras_preprocessing/image.py
@@ -2102,26 +2102,17 @@ class DataFrameIterator(Iterator):
             if not ext_exist:
                 raise ValueError('has_ext is set to True but'
                                  ' extension not found in x_col')
-            temp_df = pd.DataFrame({x_col: filenames}, dtype=str)
-            temp_df = self.df.merge(temp_df, how='right', on=x_col)
-            temp_df = temp_df.set_index(x_col)
-            temp_df = temp_df.reindex(filenames)
-            temp_df = temp_df.dropna()
-            self.filenames = list(temp_df.index)
+            self.df = self.df[self.df[x_col].isin(filenames)]
+            self.filenames = list(self.df[x_col])
         else:
             without_ext_with = {f[:-1 * (len(f.split(".")[-1]) + 1)]: f
                                 for f in filenames}
             filenames_without_ext = [f[:-1 * (len(f.split(".")[-1]) + 1)]
                                      for f in filenames]
-            temp_df = pd.DataFrame({x_col: filenames_without_ext}, dtype=str)
-            temp_df = self.df.merge(temp_df, how='right', on=x_col)
-            temp_df = temp_df.set_index(x_col)
-            temp_df = temp_df.reindex(filenames_without_ext)
-            temp_df = temp_df.dropna()
-            self.filenames = [without_ext_with[f] for f in temp_df.index]
-        self.df = temp_df.copy()
+            self.df = self.df[self.df[x_col].isin(filenames_without_ext)]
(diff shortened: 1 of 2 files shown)


keras_preprocessing/image.py
```diff
@@ -2102,26 +2102,17 @@ def __init__(self, dataframe, directory, image_data_generator,
             if not ext_exist:
                 raise ValueError('has_ext is set to True but'
                                  ' extension not found in x_col')
-            temp_df = pd.DataFrame({x_col: filenames}, dtype=str)
-            temp_df = self.df.merge(temp_df, how='right', on=x_col)
-            temp_df = temp_df.set_index(x_col)
-            temp_df = temp_df.reindex(filenames)
-            temp_df = temp_df.dropna()
-            self.filenames = list(temp_df.index)
+            self.df = self.df[self.df[x_col].isin(filenames)]
```
Vijayabhaskar96: `self.df = self.df[self.df[x_col].isin(filenames)].sort_values(by=x_col)` could you add this small change to your PR? This sorts the final dataframe and filenames.

keras_preprocessing/image.py
```diff
@@ -2102,26 +2102,17 @@ def __init__(self, dataframe, directory, image_data_generator,
...
             filenames_without_ext = [f[:-1 * (len(f.split(".")[-1]) + 1)]
                                      for f in filenames]
-            temp_df = pd.DataFrame({x_col: filenames_without_ext}, dtype=str)
-            temp_df = self.df.merge(temp_df, how='right', on=x_col)
-            temp_df = temp_df.set_index(x_col)
-            temp_df = temp_df.reindex(filenames_without_ext)
-            temp_df = temp_df.dropna()
-            self.filenames = [without_ext_with[f] for f in temp_df.index]
-        self.df = temp_df.copy()
+            self.df = self.df[self.df[x_col].isin(filenames_without_ext)]
```
Vijayabhaskar96: Same here.

Vijayabhaskar96:
@smurak Thanks for the PR, I proposed a small addition to the code please check. @Dref360 This PR looks fine to me, A tiny modification proposed to follow same filename order as `flow_from_directory`.

smurak:
@Vijayabhaskar96 Thank you for reviewing.
I've modified my codes you've pointed out for backwards compatibility.

(But I feel "shuffle=False" means DataFrameIterator generates items according to the input dataframe's order, so maybe I will submit another PR next time :slightly_smiling_face: )

Vijayabhaskar96:
@smurak I don't think 

> `shuffle=False` means DataFrameIterator generates items according to the input dataframe's order

 the filenames list is updated only after the df is sorted and also the other variables that `_get_batches_of_transformed_samples` uses. (I'm not 100% sure though)

@Dref360 This PR LGTM.

smurak:
@Vijayabhaskar96 
I mean I want DataFrameItarator to generate items according to the input dataframes's order, but it doesn't now.
I'd like to add an option "order" to DataFrameItarator to be able to choose the generating order ("directory"(current order) or "df").

Vijayabhaskar96:
I get what you're saying, You want an option something like sort_order=False so that the original order will be used. I don't think it is necessary but if you think it is useful, feel free to submit a PR (but make the sorted order the default option because the `flow_from_directory` sorts the filenames, and I think the same output format should be followed for consistency)

Issue #78, opened 2018-10-22
Off-by-one error in text preprocessing (sequence_to_matrix)


dtieber:
Seems like the num_words property in [text.py](https://github.com/keras-team/keras-preprocessing/blob/4872fccc75af22a123695dbf3b82bdb424372709/keras_preprocessing/text.py#L399) is not initialized with the correct length. I found this out because I'm using this value in order to calculate the number of input/output neurons which leads to issues when I'm training the model.

I think num_words should be initialized like this: `num_words = len(self.word_index)` if not set explicitly.

Dref360:
To avoid breaking changes, we should document this behaviour.

Dref360:
Yup I saw it. We will plan a major release with all off-by-one fixes. Right now, this would cause a lot of projects to break.

PR #79, opened 2018-10-22
Fix off-by-one error if num_words is not set explicitly 

dtieber:
### Summary
When num_words is not set explicitly on initialization of the tokenizer it is calculated on basis of the length of the dictionary when sequences_to_matrix is called. However, num_words is off by one because 1 is added to the length of the dict ( e.g. if the tokenizer is fit on the text  "car bike plane", the num_words equals 4 instead of 3).

### PR Overview

- [X] This PR is backwards compatible [y/n]

### DIFF
--- a/keras_preprocessing/text.py
+++ b/keras_preprocessing/text.py
@@ -396,7 +396,7 @@ class Tokenizer(object):
         """
         if not self.num_words:
             if self.word_index:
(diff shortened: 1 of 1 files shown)


dtieber:
Related to: [Issue#78](https://github.com/keras-team/keras-preprocessing/issues/78)

PR #82, opened 2018-10-23
Add sort and drop_duplicates option to DataFrameIterator

smurak:
### Summary

Add "sort" and "drop_duplicates" option to DataFrameIterator to be able to generate data according to just as the input dataframe order.

### Related Issues

#72

### PR Overview

### DIFF
--- a/keras_preprocessing/image.py
+++ b/keras_preprocessing/image.py
@@ -1027,7 +1027,9 @@ class ImageDataGenerator(object):
                             save_prefix='',
                             save_format='png',
                             subset=None,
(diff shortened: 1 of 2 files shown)


Vijayabhaskar96:
@smurak @Dref360 Looks good.

Dref360:
Is it right that sort and drop_duplicate are True by default? Would the user expect that?

dzhus:
It's effectively the current behaviour. It's not obvious but probably makes sense for what ImageDataGenerator would be typically used for.

Dref360:
All good LGTM

PR #83, opened 2018-10-25
Add analyzer to text preprocessing

filipecaixeta:
Add analyzer to split the text with an external function other than using only text_to_word_sequence
Includes test and documentation


Dref360:
Please add that `text_to_word_sequence` will be used otherwise.

PR #98, opened 2018-11-27
Sort classes in DataFrameIterator

mkaze:
### Summary
Currently, when the `classes` argument in `flow_from_dataframe` is not set, it would automatically find the classes by using pandas Series `unique` method. However, this method would return the unique values in order of appearance. Therefore, in scenarios where we have separate dataframes for training/validation/test it might result in inconsistent class indices mapping. Here is minimal example:

```python
import pandas as pd
import numpy as np

df_tr = pd.DataFrame({'files': ['tr_img1.jpg', 'tr_img2.jpg'],
                      'labels': ['cat', 'dog']})
df_val = pd.DataFrame({'files': ['val_img1.jpg', 'val_img2.jpg'],
                       'labels': ['dog', 'cat']})

img_gen = ImageDataGenerator()
tr_gen = img_gen.flow_from_dataframe(df_tr, 'dir', x_col='files', y_col='labels', class_mode='binary')
val_gen = img_gen.flow_from_dataframe(df_val, 'dir', x_col='files', y_col='labels', class_mode='binary')

print(tr_gen.class_indices)    # {'cat': 0, 'dog': 1}
print(val_gen.class_indices)   # {'cat': 1, 'dog': 0}
```
Notice that `cat` and `dog` are assigned different indices in train and validation generators.

To resolve this issue, this PR sort the classes to make sure the indices mapping would be the same. 
### PR Overview

### DIFF
--- a/keras_preprocessing/image.py
+++ b/keras_preprocessing/image.py
@@ -2107,7 +2107,7 @@ class DataFrameIterator(Iterator):
         if not classes:
             classes = []
             if class_mode not in ["other", "input", None]:
-                classes = list(self.df[y_col].unique())
+                classes = list(np.sort(self.df[y_col].unique()))
         else:
             if class_mode in ["other", "input", None]:
                 raise ValueError('classes cannot be set if class_mode'


mkaze:
Interesting! I think a buggy test was found out as a result of this PR. That test assumes that just setting `shuffle=True` would guarantee that the array would be in different order after shuffling. However, there is no such a guarantee. Here is a minimal example that demonstrates this:

```python
arr = np.array([1, 2, 3])

for i in range(10000):
    c_arr = np.array(arr)
    np.random.shuffle(c_arr)
    assert list(c_arr) != [1, 2, 3]
```

It is **very likely** that the above code would raise an assertion error.

Vijayabhaskar96:
@mkaze You're right, the probability of this check failing is high given that there are just 6 combinations of the shuffle is possible. @smurak Can you look into this bug in the test? @Dref360 PR looks good to me.

smurak:
@Vijayabhaskar96 I agree with you, that test can be failed with some probability.
If I fix it, set `seed` and check if the generating order is fine.

Dref360:
Please fix the test that is failling.

mkaze:
@Dref360 Fixed (#104) by setting `seed` as suggested.

Issue #102, opened 2018-12-06
valAcc disparity between flow_from_directory() and flow_from_dataframe()


Mahi-Mai:
In order to test how my model training script performed on a benchmark dataset, I converted the stored MNIST to a set of png images.  I have them organized in two ways:

**Method 1.**  I have a "train" folder and a "test" folder where images are stored without further organization.  I have a dataframe for the train and test set, with column 1 listing the absolute directory, and column 2 listing the label.  I've carefully examined this csv- the labels and image listings appear accurate.  I've captured and tested the output of flow_from_dataframe(), and it looks fine.

Sample of csv:
```
,test_samples,test_labels
0,/path/Data/test/6992.png,7
1,/path/Data/test/1380.png,9
2,/path/Data/test/5817.png,4
3,/path/Data/test/5295.png,5
4,/path/Data/test/5340.png,2
```

**Method 2.**  I have a train and test folder, each with subdirectories for the different categories of images.

Other than how they're organized, these datasets are otherwise identical.

If I run my script and use flow_from_dataframe() with the assets from Method 1, the highest validation accuracy I can manage ranges from 0.01-0.05.  If I run my script with flow_from_directory() using the assets in Method 2, my highest checkpoint is 0.93.

What could be the source of this disparity?  Am I misusing flow_from_dataframe()?  I'll share my scripts from each approach below.  Thanks in advance for any insight.

**Method 1: Garbage Validation Accuracy**
```
import pandas as pd
import numpy as np
 
import keras
from keras_preprocessing.image import ImageDataGenerator
 
from keras import applications
from keras import optimizers
from keras.models import Model 
from keras.layers import Dropout, Flatten, Dense, GlobalAveragePooling2D
from keras import backend as k 
from keras.callbacks import ModelCheckpoint, CSVLogger
 
from keras.applications.vgg16 import VGG16, preprocess_input
 
# INITIALIZE MODEL
 
img_width, img_height = 32, 32
model = VGG16(weights = 'imagenet', include_top=False, input_shape = (img_width, img_height, 3))
 
# freeze all layers
for layer in model.layers:
    layer.trainable = False
 
# Adding custom Layers 
x = model.output
x = Flatten()(x)
x = Dense(512, activation='relu')(x)
x = Dropout(0.5)(x)
predictions = Dense(10, activation="softmax")(x)
 
# creating the final model 
model_final = Model(input = model.input, output = predictions)
 
# compile the model 
[...]

Issue #99, opened 2018-11-28
Multi-task learning with DataframeIterator / flow_from_dataframe


adrianodennanni:
Hello!

I am training a image classification model with multiple outputs:
```python
trained_model = tf.keras.applications.xception.Xception(
        include_top=False,
        weights='imagenet',
        input_shape=[300, 300, 3],
        pooling='max')

    outputs = []
    for i in range(8):
      outputs.append(tf.keras.layers.Dense(1,  activation='softmax', kernel_initializer=kernel_initializer) (trained_model.output))

    model = tf.keras.Model(inputs=trained_model.input, outputs=outputs)
```

The `y` returned by this model is a Python `List`, with 8 elements. Each element is a mini-batch of tensors.

However, `flow_from_dataframe` reads all my y columns from the dataframe as one numpy array, instead of a Python `list`.
___
## Example
Suppose my dataframe is something like this:
```
image_path,field_1,field_2,field_3,field_4,field_5,field_6,field_7,field_8
1532672467738.jpeg,1,1,0,1,0,0,0,1
1532669990747.jpeg,0,0,0,1,0,1,1,0
...
```

Then I call flow_from_dataframe:
```
train_batches = generator.flow_from_dataframe(
  dataframe=dataframe,
  directory=path,
  x_col='image_path',
  y_col=['field_1', 'field_2', 'field_3', 'field_4', 'field_5', 'field_6', 'field_7', 'field_8'],
  class_mode='other',
  batch_size=16
)
```

When I call `fit_generator` with both the model and `train_batches`, I get this error:
```
ValueError: Error when checking model target: the list of Numpy arrays that you are passing to your model is not the size the model expected. Expected to see 8 array(s), but instead got the following list of 1 arrays: [array([[0, 0, 0, 0, 1, 0, 1, 1],
       [1, 1, 0, 1, 1, 0, 0, 0],
       [0, 0, 1, 0, 1, 0, 0, 1],
       [0, 0, 1, 1, 0, 0, 1, 0],
       [0, 1, 0, 1, 0, 1, 0, 1],
       [1, 0, 0, 1, 0, 1, 0, 0],
```

So, like I wrote in the beggining: `DataframeIterator` sends a numpy array of shape `(16, 8)`, and the model outputs a Python `List` of 8 numpy arrays of size `(16)`.

adrianodennanni:
[...]

Issue #95, opened 2018-11-18
ImageDataGenerator resizing and transforming images before preprocessing


ldm314:
I am trying to use the preprocessing function to take a network sized crop out of inconsistently sized input images instead of resizing to the network size. I have tried to do this using the preprocessing function but found that it is not easily possible. Using Keras 2.2.2

* ImageDataGenerator does not accept None as a type for target_size which should cause load_img to not resize things.
<pre>C:\ProgramData\Anaconda3\envs\tensorflow\lib\site-packages\keras_preprocessing\image.py in __init__(self, directory, image_data_generator, target_size, color_mode, classes, class_mode, batch_size, shuffle, seed, data_format, save_to_dir, save_prefix, save_format, follow_links, subset, interpolation)
   1665         self.directory = directory
   1666         self.image_data_generator = image_data_generator
-> 1667         self.target_size = tuple(target_size)
   1668         if color_mode not in {'rgb', 'rgba', 'grayscale'}:
   1669             raise ValueError('Invalid color mode:', color_mode,

TypeError: 'NoneType' object is not iterable</pre>
* To accomplish my goal, I modified the image data generator to pass none to load_img and then to resize afterwards if the size doesn't match the target. This hack works for my scenario:
<pre>
    def _get_batches_of_transformed_samples(self, index_array):
        batch_x = np.zeros(
            (len(index_array),) + self.image_shape,
            dtype=backend.floatx())
        # build batch of image data
        for i, j in enumerate(index_array):
            fname = self.filenames[j]
            img = load_img(os.path.join(self.directory, fname),
                           color_mode=self.color_mode,
                           target_size=None)
            x = img_to_array(img, data_format=self.data_format)
            # Pillow images should be closed after `load_img`,
            # but not PIL images.
            if hasattr(img, 'close'):
                img.close()
            params = self.image_data_generator.get_random_transform(x.shape)
            x = self.image_data_generator.apply_transform(x, params)
            x = self.image_data_generator.standardize(x)
            width_height_tuple = (self.target_size[1], self.target_size[0])
            if (x.shape[1],x.shape[0]) != width_height_tuple:
              x=cv2.resize(x,width_height_tuple, interpolation=cv2.INTER_AREA)
            batch_x[i] = x
</pre>

While looking into this I saw that the preprocessing function runs at the start of standardize, which is after the random transforms are applied. To me this sounds like preprocssing is a bad name since it isn't actually happening first.

rragundez:
[...]

Issue #88, opened 2018-11-01
flow_from_dataframe() - ValueError: has_ext is set to True but extension not found in x_col


Mahi-Mai:
I'm working with images organized across several folders.  I have a dataframe of their file paths, and up until now I've been using that with a script to move them into the necessary categorical folders.  It takes up a lot of time and space.  So, needless to say I was _ecstatic_ when I found the flow_from_dataframe() method.

I have my valid dataframe and filepaths.  I initialize the generators like this:
```
from keras.preprocessing.image import ImageDataGenerator

main_dir = '/User/name/etc/etc/224px/'

# Initiate the train and test generators with data Augumentation 
train_datagen = ImageDataGenerator(preprocessing_function = preprocess_input,
                                   #rescale = 1./255,
                                   horizontal_flip = True,
                                   fill_mode = "nearest",
                                   zoom_range = 0.3,
                                   width_shift_range = 0.1,
                                   height_shift_range = 0.1,
                                   rotation_range = 30)

test_datagen = ImageDataGenerator(preprocessing_function = preprocess_input)

seed = 108201987 # Optional random seed for shuffling and transformations.

train_generator = train_datagen.flow_from_dataframe(dataframe=train,
                                                    directory=main_dir,
                                                    x_col='filepath',
                                                    y_col='label',
                                                    has_ext=True,
                                                    target_size = (img_height, img_width),
                                                    batch_size = batch_size, 
                                                    class_mode = "binary",
                                                   seed = seed)

validation_generator = test_datagen.flow_from_directory(dataframe=val,
                                                        directory=main_dir,
                                                        x_col='filepath',
                                                        y_col='label',
                                                        has_ext=True,
                                                        target_size = (img_height, img_width),
                                                        class_mode = "binary",
                                                        seed = seed)
```
Here's a sample file path: 'October 29 2018/Top view_XYZ_4-9/IMG_6854.JPG'

Both the training and validation generators have has_ext set to True, since my files have extensions.  However, I get this error:
```
---------------------------------------------------------------------------
ValueError                                Traceback (most recent call last)
<ipython-input-11-1e162ab6ee70> in <module>()
     23                                                     batch_size = batch_size,
     24                                                     class_mode = "binary",
---> 25                                                    seed = seed)
     26 
     27 validation_generator = test_datagen.flow_from_directory(dataframe=val,

[...]

PR #107, opened 2018-12-18
Refactor image module

rragundez:
### Summary
I planned to update the image pre-processing part of Keras during my Christmas break to have a more complete cover of use-cases: multi-label, multi-outputs, dictionary iterator, perhaps others. I have faced some of this use-cases myself and have encounter people with the same use-cases, so I think is a good idea to include them in the image pre-processing module.

As I started I found that there is a single file containing all the logic with 2267 lines and all kind of different logic in there. I do not think this is good practice, and in the future (I think already now) maintaining and improving this file will be a maze. 

So to start I propose to break the file into a module, with a much better structure for iterators and the `ImageDataGenerator` containing the main logic. The classes and functions inside the `image` module can still be directly imported from `keras_preprocessing.image` as they are exposed via the `__init__.py` file. Therefore no change is required from imports depending on `keras_preprocessing.image`.
 
### Related Issues
There are no specific issues to this particular PR, as is just a refactoring, but you can already see how the quality of the code has deteriorated compared to the native Keras code, I argue that a critical part to fix bugs and contribute is to have a clean tractable structure, specially for new contributors to the project like myself. This of course will also accelerate the addition of new features on this repo to the core `Keras` repo.

For example after a quick look into the `DataframeIterator` I found many issues with it:

- Unnecessary check for pandas as it's not used anywhere
- Unnecessary check for string types of x_col, not Pythonic also problem because then what do you check and you don't check? e.g. y_col. If you want to use this you should use `typing`
- Keeps copy of the full dataframe since start even though latter it's reduced.
- Keeps x_col which is unnecessary, the iterator should only keep, classes and filenames
- What's the use-case of other? if it's in order to have continues targets then that mode should be specified. other doesn't really add much as such to the different use cases as they are very different.
- classes is numpy array while in DirectoryGenerator is a list
- Value error exception for classes which is not in DirectoryIterator. Classes can be set in any scenario, no idea why that couldn't be using `input` for example. I understand there are specific cases where indeed classes is not necessary, but the current logic does nothing to handle that. 
- Unnecessary assigning of self.classes = classes at top of init. Classes  should be unique.
- remove unnecessary check for directory None and has_ext. Directory has to be given.
- Unnecessary `ValueError` is understood that filenames contain extensions, because they are the filenames in the directory.
- For all iterators there is confusion between classes and labels, which confuses the user. Classes are unique and labels are the class for each observation.

So like i said there is no specific Issue for this change, but the below issues could be resolved faster and more accessible to new contributors with the structure change I propose. 

#99 #102 #95 #88 #67 #56 

P.D. I think you should start thinking on doing the same with `sequence.py` and `text.py` as they start growing.

### PR Overview


rragundez:
Thanks for the feedback @Dref360 , you are right about both points. I'll rollback the Mixin commit and make this PR just about the splitting of the `image` module + moving tests accordingly, then I'll open another PR with the Mixin code deduplication stuff, would be easier for you to review as well. Thanks again.

PR #118, opened 2018-12-25
Dictionary iterator

rragundez:
### Summary
This PR adds a `DictionaryIterator` class and `flow_from_dictionary` method. This class is quite flexible as it allows to send full-paths without relative to a directory parameters. Which will allow to fetch images from different root directories. The next step is to leverage this flexible API to include multi-output batches that match the functional Keras API and the multi-label one. Some of the internal logic has also been simplified in comparison with `DataframeIterator` (which in my opinion should be deprecated for several reasons, one of the main ones a dependency on the Pandas API which has substantially changed in the past).

I imagine having the well known `DirectoryIterator` for the already known simple use cases and `DictionaryIterator` for the more advance ones because of flexibility.

### PR Overview

### DIFF
--- a/keras_preprocessing/image/dataframe_iterator.py
+++ b/keras_preprocessing/image/dataframe_iterator.py
@@ -21,17 +21,12 @@ class DataFrameIterator(Iterator):
 
     # Arguments
         dataframe: Pandas dataframe containing the filepaths relative to
-            `directory` of the images in a column and classes in another
-            column/s that can be fed as raw target data.
-        directory: Path to the directory to read images from.
-            Each subdirectory in this directory will be
-            considered to contain images from one class,
-            or alternatively you could specify class subdirectories
-            via the `classes` argument.
-            if used with dataframe,this will be the directory to under which
-            all the images are present.
-            You could also set it to None if data in x_col column are
-            absolute paths.
+            `directory` or absolute paths if `directory` is None of the images
+            in a column and classes in another column/s that can be fed as raw
+            target data.
+        directory: string, path to the directory to read images from. Directory to
+            under which all the images are present. If None, data in x_col column
+            should be absolute paths.
         image_data_generator: Instance of `ImageDataGenerator` to use for
             random transformations and normalization. If None, no transformations
             and normalizations are made.
--- a/keras_preprocessing/image/dictionary_iterator.py
+++ b/keras_preprocessing/image/dictionary_iterator.py
@@ -56,7 +56,9 @@ class DictionaryIterator(Iterator):
             By default, "nearest" is used.
         dtype: Dtype to use for generated arrays.
     """
-    allowed_class_modes = {'categorical', 'binary', 'sparse', 'input', None}
+    allowed_class_modes = {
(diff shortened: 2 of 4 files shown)


Dref360:
I think we will stick with the Pandas API because it's widely used in the scientific community. If their API does changes, we will set a hard dependency version.

Since the `flow_from_dataframe` is pretty flexible already, I don't think we should merge your PR.

rragundez:
@Dref360 OK I understand your point. I want to point out that normally when a Pandas dependency is added to an ML library is as an add-on feature. It normally first implemented with native Python code and data structures as the base, only after the add-on to support an external widely use library like Pandas is added, but rarely for mature library is the default way to go to use an external library. You can see that in the scikit packages for example. In general I don't think is a good practice to leave your core functionality to an external library.  For example I do not see a Pandas dependency in the complete Keras source code (no import but also no `DataFrame` method) but maybe I'm wrong. So I do not know the stands of @fchollet on this but perhaps it won't be easy to merge `DataFrameIterator` into the Keras source code. 

Have said that I understand your point and I'll concentrate then on updating the code in the DataframeIterator. Thanks for the feedback. Hopefully I can get to the multi-label and multi-output use cases before new years. Thanks again.

Dref360:
1. Pandas is used inside keras codebase as a soft deps.
 i. https://github.com/keras-team/keras/blob/0cfa5c2709906a7a76f552f71a562f899e408695/keras/engine/training_utils.py#L33

2. `but perhaps it won't be easy to merge DataFrameIterator into the Keras source code`
 i. I'm not sure to get what you mean. keras-preprocessing is a standalone package that won't get merged into Keras.

As for this PR, I don't think we will merge it because the DataframeIterator should be the "standard" way of handling anything that cannot be done with flow_from_directory.

I would still like @fchollet input on that feature to be sure.

rragundez:
I mean merged into: https://github.com/keras-team/keras/blob/master/keras/preprocessing/image.py like `DirectoryIterator` and `ImageDataGenerator`. I guess the integration there is the final step for a mature implementation. 

I agree with you @Dref360 that there should be preferably one way to deal with something that is not done by the `flow_from_directory`. So I started efforts to update `DataFrameIterator`.

Dref360:
We should revisit the file you linked, I don't think it's useful anymore.

________________________________
De : rragundez <notifications@github.com>
Envoyé : 30 décembre 2018 12:01:33
À : keras-team/keras-preprocessing
Cc : Frédéric Branchaud-Charron; Mention
Objet : Re: [keras-team/keras-preprocessing] Dictionary iterator (#118)


I mean merged into: https://github.com/keras-team/keras/blob/master/keras/preprocessing/image.py like DirectoryIterator and ImageDataGenerator. I guess the integration there is the final step for a mature implementation.

I agree with you @Dref360<https://github.com/Dref360> that there should be preferably one way to deal with something that is not done by the flow_from_directory. So I started efforts to update DataFrameIterator.

rragundez:
I've added multi-ouput and multi-label support

rragundez:
Since we are doing a lot of work on the `DataFrameIterator` which this PR added I think this PR feature is not relevant anymore.

PR #119, opened 2018-12-25
Mode class modes as class attribute and refactor error message

rragundez:
### Summary
Class modes are currently hardcoded in the code logic, and present in two places. In the logic check and in the error message. This PR solves both things, it's not a good practice to hardcode in the logic and if not possible then is best to make it an attribute, then this PR also refactor the error message to use this attribute such that if it's updated the error message updates as well.

### PR Overview


Dref360:
LGTM once the conflicts are resolved.

rragundez:
@Dref360 conflicts are resolved now.

PR #122, opened 2018-12-26
Remove sort

rragundez:
### Summary
Currently the `DataframeIterator` has an argument `sort`. This argument doesn't add to the internal logic that `DataframeIterator` should worry about. Even if used, `sort` is used before `shuffle` for yielding the batches so there is no use for it other than having a pretty arrangement of the `filenames` attribute. But if that's what the user wants then the user can call `sort_values` on the dataframe column before sending it as an argument. This PR removes the `sort` logic which it actually only calls `sort_values` form Pandas, this improves readability in the code, easier to maintain as the logic was in several places and also in my opinion sorting has nothing to do with the logic that  `DataframeIterator` should be handling.

### PR Overview

### DIFF
--- a/keras_preprocessing/image/dataframe_iterator.py
+++ b/keras_preprocessing/image/dataframe_iterator.py
@@ -72,7 +72,6 @@ class DataFrameIterator(Iterator):
             If PIL version 1.1.3 or newer is installed, "lanczos" is also
             supported. If PIL version 3.4.0 or newer is installed, "box" and
             "hamming" are also supported. By default, "nearest" is used.
-        sort: Boolean, whether to sort dataframe by filename (before shuffle).
         drop_duplicates: Boolean, whether to drop duplicate rows based on filename.
     """
     allowed_class_modes = {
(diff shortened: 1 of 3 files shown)


Vijayabhaskar96:
> other than having a pretty arrangement of the filenames attribute.
 
Yes, that's exactly why sort was included, it can be removed too to keep things simple, I remember `DirectoryIterator` had it's filename atrribute which was sorted that is the reason `sort `as included here to maintain same format as `DirectoryIterator`, since it's no longer the case, it can be removed from here too.
@Dref360 PR looks good to me, but may need Deprication warning message to be added.

rragundez:
I think I will add the warning indeed, and also a test to make sure the batches of files, if shuffle is set to False are returned in the order they are provided. thanks!

rragundez:
@Vijayabhaskar96 I think is done, please have a look.

Vijayabhaskar96:
@rragundez Looks fine, Thanks! @Dref360

Dref360:
LGTM thanks

PR #123, opened 2018-12-26
Remove has ext

rragundez:
### Summary
At the moment there is in my view unnecessary logic in the `DataframeIterator` class to handle user provided filenames without extension. Also this unnecessary logic leads to bad code practices and bugs. Some of them are:
- unnecessary double check for valid list in directory, as it only works if directory is set (137-144 and 149) it  can be done in one pass.
- logic for has_ext only checks the first element of x_col. If first element has extension
but all the other do not then the logic fails (line 153)
- why has_ext cannot be set to False if directory is none? the user should give then the fullpaths in the dataframe (lines 145-148). This logic fails because you have basically two places defining if there is an extensions (the has_ext and the filenames in the dataframe themselves).
- then such a complicated logic to deal with file extensions if has_ext is false (lines 160-164). 

The committer mentions the use case of a Kaggle competition in (https://medium.com/@vijayabhaskar96/tutorial-on-keras-flow-from-dataframe-1fd4493d237c), but in the use cases I've worked in real-life I've never received the data like this (images in disk, some collection of filenames without extension). Normally we collect the filenames and then (as he points out in the blog post) I build my own `DataGenerator` which extends from `Sequence` where I feed in this list of filenames pointing to some file on disk, this is the most common use case by far and I've seen it several times. I think this library aims to solve real use cases.

In short I think the user should be responsible for setting filenames in the `x_col` which exist in the `directory`. This will also allow the user to have flexibility if some have extension and other don't for example.

This PR simplifies the whole process of getting valid filenames, basically doing the computation in a single pass of the df by filtering it. It also cleans up a bit of unnecessary attributes being set which later are overwritten.

### PR Overview

### DIFF
--- a/keras_preprocessing/image/utils.py
+++ b/keras_preprocessing/image/utils.py
@@ -177,14 +177,10 @@ def _list_valid_filenames_in_directory(directory, white_list_formats, split,
         follow_links: boolean.
 
     # Returns
-        classes: a list of class indices(returns only if `df=False`)
-        filenames: if `df=False`,returns the path of valid files in `directory`,
-            relative from `directory`'s parent (e.g., if `directory` is
-            "dataset", the filenames will be
-            `["dataset/class1/file1.jpg", "dataset/class1/file2.jpg", ...]`).
-            if `df=True`, returns the path of valid files in `directory`,
-            relative from `directory` (e.g., if `directory` is
-            "dataset", the filenames will be
+         classes: a list of class indices
+         filenames: the path of valid files in `directory`, relative from
+             `directory`'s parent (e.g., if `directory` is "dataset/class1",
+            the filenames will be
             `["class1/file1.jpg", "class1/file2.jpg", ...]`).
     """
     dirname = os.path.basename(directory)
--- a/keras_preprocessing/image/utils.py
+++ b/keras_preprocessing/image/utils.py
@@ -160,7 +160,7 @@ def _iter_valid_files(directory, white_list_formats, follow_links):
 
 
 def _list_valid_filenames_in_directory(directory, white_list_formats, split,
-                                       class_indices, follow_links, df=False):
+                                       class_indices, follow_links):
     """Lists paths of files in `subdir` with extensions in `white_list_formats`.
 
     # Arguments
@@ -175,7 +175,6 @@ def _list_valid_filenames_in_directory(directory, white_list_formats, split,
             of images in each directory.
         class_indices: dictionary mapping a class name to its index.
         follow_links: boolean.
-        df: boolean
 
     # Returns
         classes: a list of class indices(returns only if `df=False`)
@@ -199,13 +198,6 @@ def _list_valid_filenames_in_directory(directory, white_list_formats, split,
     else:
         valid_files = _iter_valid_files(
             directory, white_list_formats, follow_links)
-    if df:
-        filenames = []
-        for root, fname in valid_files:
-            absolute_path = os.path.join(root, fname)
-            relative_path = os.path.relpath(absolute_path, directory)
-            filenames.append(relative_path)
-        return filenames
     classes = []
     filenames = []
     for root, fname in valid_files:
--- a/keras_preprocessing/image/dataframe_iterator.py
+++ b/keras_preprocessing/image/dataframe_iterator.py
@@ -143,14 +143,13 @@ class DataFrameIterator(Iterator):
         self.df = self._filter_valid_filepaths(self.df)
         if sort:
             self.df.sort_values(by=x_col, inplace=True)
-        self.filenames = self.df[x_col].tolist()
 
         if self.split:
-            num_files = len(self.filenames)
+            num_files = len(self.df)
             start = int(self.split[0] * num_files)
             stop = int(self.split[1] * num_files)
             self.df = self.df.iloc[start: stop, :]
-            self.filenames = self.filenames[start: stop]
+        self.filenames = self.df[x_col].tolist()
 
         if class_mode not in ["other", "input", None]:
             classes = self.df[y_col].values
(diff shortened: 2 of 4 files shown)


Vijayabhaskar96:
I don't think there is a need to remove has_ext, considering removing it breaks backward compatibility.I agree It may need better implementation, but  I do think has_ext is useful, even in mixed cases where you have filenames both with and without mixed together in a dataframe, the user can manually update the filenames in dataframe and set ` has_ext=True` or remove the extensions if it exists and set `has_ext=False.` @smurak @Dref360 your views?

rragundez:
@Vijayabhaskar96  thanks for the comments. I hope you don't take my comments wrongly as I do want to help out build a solid Keras preprocessing code base, just pointing some constructive criticism.
- "breaks backward compatibility" that is indeed a drawback, but this cannot be the only reason by itself, specially since this implementation is relatively new so I think is better to discuss now.
- "but I do think has_ext is useful", many things are useful but that's not why features are implemented in software. I haven't seen in my experience a real use case where this `has_ext` would be helpful with the purpose of providing sort of filenames IDs.
- "the user can manually update the filenames in dataframe and set `has_ext=True`", this means that the user had already encounter the problematic logic and probably already lost some time debugging, and if you're asking the user to already modify the filenames in the dataframe, how is that different from asking to the user clearly since the beginning that filenames in the dataframe had to match does on disk?

Unfortunately by design this problematic logic has no solution because there are two places defining if there are extensions (the has_ext and the filenames in the dataframe themselves). The alternative is to remove `has_ext` and then if a filename in a dataframe has no extension then we look in the directory a matching image file with that filename regardless of the extension. This of course has also its own problems and a over-complicated logic.Therefore I think is better to remove it before the code base becomes more complicated when adding further features.

Vijayabhaskar96:
I do understand your point, I didn't take anything wrong from your comments, I will be happy to see this function work effectively with your PR merged but I'm not sure removing `has_ext` completely is a good choice or not, I created this function for myself and pull requested here because I thought this may help others too. Since I made this for myself at the beginning I didn't make it simple and efficient, it does needs changes and I don't have much experience here, infact this was my first major open source contribution, that is why asked others for their view on it,

rragundez:
@Vijayabhaskar96 ah well the DataframeIterator is certainly helpful specially to cover other use cases outside `DirectoryIterator`, so thanks for that. I do think we need to have a critical look at it because this class can become the basis for handling all other use cases (multi-ouput, multi-label, regression, etc.). Let's see what the main committers think about it. In my opinion we need to clean it a bit, make it very concise and move forward with other features.

smurak:
For me, I don't feel the necessity of the `has_ext` option.
If images have same file extension, we can easily insert the extension by pandas' `map` method,
and I think extension mismatch case is unlikely to occur when we are given only filenames (without extensions) in dataframes.
So for simplicity, I think the `has_ext` option could be removed.

This PR looks good, but I have a concern.
I remember `_list_valid_filepaths` is slower than `_list_valid_filenames_in_directory`,
so without `_list_valid_filenames_in_directory`, the code become simpler but the performance worse.
I'm not sure this performance degradation is acceptable.

rragundez:
thanks for the comments @smurak, indeed `_list_valid_filepaths` was probably slow but I am not using that i replaced it with `_filter_valid_filepaths` which filters the dataframe provided in one pass instead of getting a collection of filepaths and then filtering on the dataframe, which  expect was a slow 2 step process. But in any case even you will use `_list_valid_filenames_in_directory` you will still have that 2 step process because you still need to cross reference with the filenames given by the user in the DataFrame. 
So as a whole I expect the PR flow to be faster because of the single step process. Please let me know what you think? If you are OK with it please merge and I will proceed rebasing the other changes. Thanks

Vijayabhaskar96:
Indeed I believe your single step process is better than `_list_valid_filenames_in_directory `, speaking of which the `_list_valid_filenames_in_directory` was made to handle both `DirectoryIterator` and `DataframeIterator` So I think it needs removal of `df=False` to make this PR complete.

Vijayabhaskar96:
@rragundez That was quick! but you forgot to update docs.  @Dref360 PR Looks good to me.

rragundez:
which docs? can you point me to them? @Vijayabhaskar96

keras_preprocessing/image/utils.py
```diff
@@ -175,7 +175,6 @@ def _list_valid_filenames_in_directory(directory, white_list_formats, split,
             of images in each directory.
         class_indices: dictionary mapping a class name to its index.
         follow_links: boolean.
-        df: boolean
 
     # Returns
         classes: a list of class indices(returns only if `df=False`)
```
Vijayabhaskar96: Here

rragundez: thanks. Done

Vijayabhaskar96: Good. Thanks for the PR!

smurak:
@rragundez 
> i replaced it with _filter_valid_filepaths

Thanks for the notice.
I just checked the performance again and  didn't find any degradation.
Looks good to me.

> If you are OK with it please merge

I'm not a committer.
@Dref360  This PR is not backward compatible, but looks good to me.

Dref360:
I don't mind breaking changes in private methods, but the `flow_from_dataframe` call should stay the same imo.

Could we deprecate has_ext for 1 minor?

 We do not have a clear deprecation procedure. I'll hit up other Keras contributors and try to start it.

Dref360:
In this case, we need to raise a DeprecationWarning when use_ext is set to false. I'll merge after.

rragundez:
@Dref360 done! :)

Dref360:
LGTM thanks

PR #127, opened 2018-12-29
[WIP] Update base Iterator class

rragundez:
### Summary
Currently both `DataFrameIterator` and `DirectoryIterator` contain the `_get_batches_of_transformed_samples` method, but both methods do the same thing, and the logic is slightly different. This PR merges both method versions and adds the method to the base class. This PR avoid this code duplication and in my opinion also simplifies the logic a little bit by stating that is enough to collect `filepaths` and `labels` from any child class of `Iterator` to make things work. Therefore both `filepaths` and `labels` are declared as properties and an `NotImplementedError` is raised if not overwritten by the child class.

### PR Overview

### DIFF
--- a/keras_preprocessing/image/dataframe_iterator.py
+++ b/keras_preprocessing/image/dataframe_iterator.py
@@ -147,7 +147,7 @@ class DataFrameIterator(Iterator):
             classes = self.df[y_col].values
             self.classes = np.array([self.class_indices[cls] for cls in classes])
         elif class_mode == "other":
-            self.data = self.df[y_col].values
+            self._data = self.df[y_col].values
             if type(y_col) == str:
                 y_col = [y_col]
             if "object" in list(self.df[y_col].dtypes):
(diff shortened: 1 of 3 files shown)


Dref360:
I'll be able to review only on Sunday. Sorry for the delay.

Vijayabhaskar96:
> Currently both DataFrameIterator and DirectoryIterator contain the _get_batches_of_transformed_samples method, but both methods do the same thing, and the logic is slightly different
Yes, they are almost the same. In fact, I suggested the same, to merge them both when `DataframeIterator `was first accepted, but others thought that it is best to leave them separately so that each Iterator can be maintained separately. Currently, if you merge them both, the Numpy Iterator will be the only one which implements its own `_get_batches_of_transformed_samples` and any other future Iterator would have to write their own too if needed, that will make only the `_get_batches_of_transformed_samples` function of `DataFrameIterator` and `DirectoryIterator` reside inside the Iterator, which doesn't seem to be a good choice to me, even though they have redundant code.

rragundez:
That's a very good point. I'll change it to a MixIn, that should solve the problem.

rragundez:
@Vijayabhaskar96 I see where the complication comes about. There is a method called `common_init` in the base `Iterator` class which is not really common as it's not used by the `NumpyArrayIterator` (also no idea why is it there as it does nothing related to what an iterator does which is the class itself), so then the `_get_batches_of_transformed_samples` cannot be in the in the same class because of what you pointed out but it cannot be in a Mixin neither because it uses attributes set in the `common_init`. Then the best practice is to move `_get_batches_of_transformed_samples` and `common_init` to a separate Mixin which is only use by the `DataframeIterator` and the `DirectoryIterator`. 

This work is far more than what this PR was made for. I'll close this one and open another one once I made the changes I mentioned above.

PR #130, opened 2019-01-05
Refactor iterator

rragundez:
### Summary
Currently both `DataFrameIterator` and `DirectoryIterator` contain the `_get_batches_of_transformed_samples` and `common_init` method, but both classes methods do the same thing, and the logic is slightly different. This PR merges both method versions and adds them to a Mixin class `BatchFromFilesMixin`. It needs to be a Mixin and not in the `Iterator` class because `NumpyIterator` does not use them. This PR avoid this code duplication and in my opinion also simplifies the logic a little bit by stating that is enough to collect `filepaths` and `labels` from to make things work, which makes sense. Therefore both `filepaths` and `labels` are declared as properties and an `NotImplementedError` is raised if not overwritten by the child class.

@Dref360, After @Vijayabhaskar96 pointed out some problems with PR #127 so I closed it. This PR does what #127 was doing plus some.

### PR Overview

### DIFF
--- a/keras_preprocessing/image/dataframe_iterator.py
+++ b/keras_preprocessing/image/dataframe_iterator.py
@@ -7,14 +7,14 @@ from __future__ import print_function
 import os
 import numpy as np
 
-from .iterator import Iterator
+from .iterator import BatchFromFilesMixin, Iterator
 from .utils import (array_to_img,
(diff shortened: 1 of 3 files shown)


rragundez:
Any update on this one?

Issue #135, opened 2019-01-09
Does flow_from_dataframe() support multilabel outputs?


scott-zockoll:
I have previously been able to train a CNN to have multi-class and multi-label output using flow(), but I am having trouble getting it to train where the labels are vectors of binary values. I am not sure what class_mode I should be using? @Vijayabhaskar96 is this supported with flow_from_dataframe()?

rragundez:
@scott-zockoll working on it today :)

rragundez:
there you go https://github.com/keras-team/keras-preprocessing/pull/136
Hopefully it will get merge soon. A +1 is always appreciated in the PR.

Vijayabhaskar96:
@scott-zockoll Sorry I was busy didn't notice the issue.
I have yet to see @rragundez 's PR You could change the multiple classes into one-hot vectors and split each column into separate columns such that the dataframe looks like this:
```
|filenames|class1|class2|class3|
|1.png|0|1|1|
|2.png|0|1|0|
|3.png|1|1|1|
...
```
You can perform this easily with Sklearn's [MultiLabelBinarizer](https://scikit-learn.org/stable/modules/generated/sklearn.preprocessing.MultiLabelBinarizer.html)

You can now pass the `x_col="filenames" , y_col=["class1","class2","class3"] and class_mode="other"`
but this method only allows you to use a single Dense layer with 3 (no. of classes) neurons in the last layer with single loss.
If you want to use multiple loss functions you may want to write a python generator function that loops through the output of flow_from_dataframe and yields the reshaped array,this generator can be used now to fit the model with multiple losses.

rragundez:
@Vijayabhaskar96 indeed I think that would work. Just something I wanted to point out:

By using the MultiLabelBinarizer or basically expanding the tools you impact the memory footprint of the algorithm without necessity, not only during the expansion process but also the class it self `DataFrameIterator` will have to keep in memory the full expanded matrix or dataframe of values for each observation and class. In many realistic scenarios this is actually quite a lot since many times you have tens or hundreds of classes and  thousand examples easily. In the PR this is optimize by keeping the minimum information necessary.

An option would be to support SparseDataFrame BTW, perhaps too far fetch.

scott-zockoll:
@rragundez Great. I will check our your PR.

[...]

PR #136, opened 2019-01-11
Multi label

rragundez:
### Summary
At the moment there is no support for multi-label. This PR adds multi-label support.
Minimal example:
```
import numpy as np
import scipy.misc
import pandas as pd
import random
import matplotlib.pyplot as plt

pixel_val = 1
filenames = []
for i in range(30):
    filename = '/tmp/{}.jpg'.format(i)
    plt.imsave(filename, pixel_val * np.ones((3, 3, 3)))
    filenames.append(filename)

classes = ['dog', 'cat', ['dog', 'cat']] * 10
df = pd.DataFrame({'filename': filenames, 'class': classes}).sample(frac=1).reset_index(drop=True)

generator = ImageDataGenerator(rescale=1/255).flow_from_dataframe(
    df,
    batch_size=3,
    target_size=(3, 3),
    shuffle = False,
    classes=['non-existent', 'dog', 'cat'],
    class_mode='categorical',
)
assert next(generator)[1].shape == (3, 3)
```
The way to use it integrates seemingly to the current way of using it. The user can still use the `categorical` class mode but instead send several classes per observation in a form of a list or tuple. It will just work.

There is indeed some additional boiler plate for checking inputs and parameters, there were already some before but now they have moved to the top of the function and to a separate method logic.

### Related Issues
#135 

### PR Overview

### DIFF
--- a/keras_preprocessing/image/dataframe_iterator.py
+++ b/keras_preprocessing/image/dataframe_iterator.py
@@ -49,16 +49,19 @@ class DataFrameIterator(BatchFromFilesMixin, Iterator):
             Color mode to read images.
         classes: Optional list of strings, classes to use (e.g. `["dogs", "cats"]`).
             If None, all classes in `y_col` will be used.
-        class_mode: Mode for yielding the targets:
-            `"binary"`: binary targets (if there are only two classes),
-            `"categorical"`: categorical targets, supports multi-label output
-                if class values per observation are stored in a list or tuple
-                in the `y_col` column,
-            `"sparse"`: integer targets,
-            `"input"`: targets are images identical to input images (mainly
-                used to work with autoencoders),
-            `"other"`: targets are the data(numpy array) of y_col data
-            `None`: no targets get yielded (only input images are yielded).
+        class_mode: one of "categorical", "binary", "sparse", "input",
+            "other" or None. Default: "categorical".
+            Mode for yielding the targets:
+            - `"binary"`: 1D numpy array of binary labels,
+            - `"categorical"`: 2D numpy array of one-hot encoded labels.
+                Supports multi-label output.
+            - `"sparse"`: 1D numpy array of integer labels,
+            - `"input"`: images identical to input images (mainly used to
+                work with autoencoders),
+            - `"other"`: numpy array of `y_col` data,
+            - `None`, no targets are returned (the generator will only yield
+                batches of image data, which is useful to use in
+            `model.predict_generator()`, `model.evaluate_generator()`, etc.).
         batch_size: Integer, size of a batch.
         shuffle: Boolean, whether to shuffle the data between epochs.
         seed: Random seed for data shuffling.
--- a/keras_preprocessing/image/dataframe_iterator.py
+++ b/keras_preprocessing/image/dataframe_iterator.py
@@ -170,8 +170,7 @@ class DataFrameIterator(BatchFromFilesMixin, Iterator):
         if not all(df[x_col].apply(lambda x: isinstance(x, str))):
             raise ValueError('All values in column x_col={} must be strings.'
                              .format(x_col))
-        # check labels are string or numeric if class_mode is binary or sparse
-        # Note: categorical class_mode supports numeric, string, list and tuple
+        # check labels are string if class_mode is binary or sparse
         if self.class_mode in {'binary', 'sparse'}:
             if not all(df[y_col].apply(lambda x: isinstance(x, str))):
                 raise TypeError('If class_mode="{}", y_col="{}" column '
@@ -188,14 +187,14 @@ class DataFrameIterator(BatchFromFilesMixin, Iterator):
             elif df[y_col].nunique() != 2:
                 raise ValueError('If class_mode="binary" there must be 2 classes. '
                                  'Found {} classes.'.format(df[y_col].nunique()))
-
+        # check values are string, list or tuple if class_mode is categorical
         if self.class_mode == 'categorical':
             types = (str, list, tuple)
             if not all(df[y_col].apply(lambda x: isinstance(x, types))):
                 raise TypeError('If class_mode="{}", y_col="{}" column '
                                 'values must be type string, list or tuple.'
                                 .format(self.class_mode, y_col))
-        # check that no classes are given if class_mode other or input
+        # raise warning if classes are given and class_mode other or input
         if classes and self.class_mode in {"other", "input", None}:
             warnings.warn('`classes` will be ignored given the class_mode="{}"'
                           .format(self.class_mode))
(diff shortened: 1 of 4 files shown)


Vijayabhaskar96:
@Dref360 PR Looks good to me. @rragundez Thanks! Tested and works fine.

rragundez:
I'll update the documentation a bit more and then merge it, thanks!

keras_preprocessing/image/dataframe_iterator.py
```diff
@@ -192,23 +255,22 @@ def _get_batches_of_transformed_samples(self, index_array):
...
+            for i, n_observation in enumerate(index_array):
+                batch_y[i, self.classes[n_observation]] = 1.
         elif self.class_mode == 'other':
             batch_y = self.data[index_array]
         else:
             return batch_x
         return batch_x, batch_y
 
-    def _filter_valid_filepaths(self, df):
+    def _filter_valid_filepaths(self, df, x_col):
```
Dref360: No doc for `x_col`

tests/image_test.py
```diff
@@ -638,6 +638,52 @@ def test_dataframe_iterator_class_mode_input(self, tmpdir):
...
         assert(input_img[0][0][0] != output_img[0][0][0])
 
+    def test_dataframe_iterator_class_mode_categorical_multi_label(self, tmpdir):
+        # save the images in the paths
+        filenames = []
+        count = 0
+        for test_images in self.all_test_images:
+            for im in test_images:
+                filename = 'image-{}.png'.format(count)
+                im.save(str(tmpdir / filename))
```
Dref360: please delete the files at the end of the test.

keras_preprocessing/image/dataframe_iterator.py
rragundez: done

keras_preprocessing/image/dataframe_iterator.py
```diff
@@ -49,16 +49,19 @@ class DataFrameIterator(Iterator):
...
+            - `"binary"`: 1D numpy array of binary labels,
+            - `"categorical"`: 2D numpy array of one-hot encoded labels.
+                Supports multi-label output.
+            - `"sparse"`: 1D numpy array of integer labels,
+            - `"input"`: images identical to input images (mainly used to
+                work with autoencoders),
+            - `"other"`: numpy array of `y_col` data,
+            - `None`, no targets are returned (the generator will only yield
+                batches of image data, which is useful to use in
+            `model.predict_generator()`, `model.evaluate_generator()`, etc.).
```
Vijayabhaskar96: `model.evaluate_generator()` takes both x and y data and returns loss by comparing it with predictions, So I don't think it will work with `class_mode=None`

rragundez: are you sure? because I also see it in `fir_from_directory` docstring

https://github.com/keras-team/keras-preprocessing/blame/master/keras_preprocessing/image/image_data_generator.py#L484

Vijayabhaskar96: Evaluate generator does need y data, refer the while loop statement here(https://github.com/keras-team/keras/blob/1336cdb14ff03de754aec6899794742ca91057b2/keras/engine/training_generator.py#L346)
If class_mode=None, y data is not returned right?

rragundez: You are correct. Thanks for pointed it out. I'll make the changes in both places.

rragundez: is it OK now? please let me know if you notice something else.

Vijayabhaskar96:
Looks good! @rragundez Thanks! @Dref360

keras_preprocessing/image/dataframe_iterator.py
```diff
@@ -20,32 +23,45 @@ class DataFrameIterator(Iterator):
...
+            depending on the `class_mode`:
+                - if `class_mode` is `"categorical"` (default value) it must
+                include the `y_col` column with the class/es of each image.
+                Values in column can be string/list/tuple if a single class
+                or list/tuple if multiple classes.
+                - if `class_mode` is `"binary"` or `"sparse"` it must include
+                the given `y_col` column with class values as strings.
+                - if `class_mode` is `"other"` it should contain the columns
+                specified in `y_col`.
+                - if `class_mode` is `input` or None no extra column is needed.
```
Dref360: missing some " and `

keras_preprocessing/image/image_data_generator.py
```diff
@@ -565,9 +564,23 @@ def flow_from_dataframe(self,
...
+                    - if `class_mode` is `"binary"` or `"sparse"` it must include
+                    the given `y_col` column with class values as strings.
+                    - if `class_mode` is `"other"` it should contain the columns
+                    specified in `y_col`.
+                    - if `class_mode` is `input` or None no extra column is needed.
             directory: string, path to the directory to read images from.
-                Directory to under which all the images are present.
-                If None, data in x_col column should be absolute paths.
+                Directory to under which all the images are present. If None,
+                data in `x_col` column should be absolute paths. If None,
```
Dref360: duplicated doc

keras_preprocessing/image/dataframe_iterator.py
```diff
@@ -102,61 +118,127 @@ def __init__(self,
...
+    def _check_params(self, df, x_col, y_col, classes):
+        # check class mode is one of the currently supported
+        if self.class_mode not in self.allowed_class_modes:
+            raise ValueError('Invalid class_mode: {}; expected one of: {}'
+                             .format(self.class_mode, self.allowed_class_modes))
+        # check that filenames/filepaths column values are all strings
+        if not all(df[x_col].apply(lambda x: isinstance(x, str))):
+            raise ValueError('All values in column x_col={} must be strings.'
+                             .format(x_col))
+        # check labels are string or numeric if class_mode is binary or sparse
```
Dref360: the comment is not right I think. You only check for strings, not numeric.

Dref360: I think it does make sense to support numeric.

Vijayabhaskar96: Since the has_ext option has been removed, It expects file extensions. So doesn't support numeric now.

keras_preprocessing/image/dataframe_iterator.py
rragundez: fixed

keras_preprocessing/image/image_data_generator.py
rragundez: good catch! Fixed.

keras_preprocessing/image/dataframe_iterator.py
rragundez: The `x_col` are just the filenames. Sort of obvious check but the feedback to the user with the message is the nice thing here. Since it mentions explicitly the column being used.

Issue #85, opened 2018-10-26
ImageDataGenerators ``standardize`` modifies input


alxy:
I am not sure if this is a bug or intended functionality, but I realized that ``ImageDataGenerator``s ``standardize()`` method is actually modifying the input passed to it. This can be quite annoying in the context of jupyter notebook, when you want to reuse the raw images again. Consider the following example code:

```python
from keras.preprocessing.image import ImageDataGenerator

datagen = ImageDataGenerator(rescale=2)

images = np.ones((10, 128, 128, 3))
print(images.mean()) # 1.0

images_std = datagen.standardize(images)

print(images.mean(), images_std.mean()) # 2.0 2.0
```

I tracked this down to the fact, that this function uses the shorthand operator for simple arithmetic operations: https://github.com/keras-team/keras-preprocessing/blob/master/keras_preprocessing/image.py#L1124 The following two functions do exactly the same thing, however the first one does not modify the input while the second does:

```python
def multiply1(x):
    x = x * 2

def multiply2(x):
    x *= 2
    
    
images = np.ones((10, 128, 128, 3))
print(images.mean()) # 1.0

multiply1(images)
print(images.mean()) # 1.0

multiply2(images)
print(images.mean()) # 2.0
```

Note that the behavior also differs for different standardization steps, e.g. ``zca_whitening`` does not modify the input while ``rescale`` does.

Dref360:
It is as designed, mostly to reuse memory when possible. When using jupyter, I would suggest providing a copy of the input instead.

rragundez:
In this case I agree with @alxy. This is a bad practice and should be avoided, as it is an unexpected behavior that faces the user and can have unexpected consequences. Since Keras puts the user first this should be avoided on any function facing the user directly. The behavior of course reduces the memory footprint and can bring benefits, but it should only be used on internal logic not on top-level user facing functions.

That's why for example we do this: https://github.com/keras-team/keras-preprocessing/blob/master/keras_preprocessing/image/dataframe_iterator.py#L106 even though this holds two copies of the dataframe at some point in time, but the user is first so we do it.

I made a PR #138 to address this issue. If there are more places where this happens it should be addressed as well.

gglanzani:
Modify in place wouldn't be bad if you wouldn't return a value.

If the user wishes to save memory, passing an extra param (in_place for example) is better suited.

rragundez:
[...]

PR #138, opened 2019-01-12
Refactor standardize

rragundez:
### Summary
At the moment the `standardize` method of `ImageDataGenerator` change the input inplace. This is a bad practice and should be avoided, as it is an unexpected behavior that faces the user and is unexpected. Since Keras puts the user first this should be avoided on any function facing the user directly. The behavior of course reduces the memory footprint and can bring benefits, but it should only be used on internal logic not on top-level user facing functions. 

This PR changes the behavior of the function to not change the user's input. It also adds tests to check that the input is not changed.

```

datagen = ImageDataGenerator(rescale=2)

images = np.ones((10, 128, 128, 3))
print(images.mean()) # 1.0

images_std = datagen.standardize(images)

print(images.mean(), images_std.mean()) # 2.0 2.0
```
### Related Issues
#85 

### PR Overview

### DIFF
--- a/keras_preprocessing/image/image_data_generator.py
+++ b/keras_preprocessing/image/image_data_generator.py
@@ -668,7 +668,15 @@ class ImageDataGenerator(object):
         )
 
     def standardize(self, x):
-        """Applies the normalization configuration to a batch of inputs.
+        """Applies the normalization configuration in-place to a batch of inputs.
+
+        `x` is changed in-place since the function is mainly used internally
+        to standarize images and feed them to your network. If a copy of `x`
+        would be created instead it would have a significant performance cost.
+        If you want to apply this method without changing the input in-place
+        you can call the method creating a copy before:
+
+        standarize(np.copy(x))
 
         # Arguments
             x: Batch of inputs to be normalized.


rragundez:
I aded the `inplace` parameter as discussed in the issue. What do you think now? @gglanzani  @alyx

keras_preprocessing/image/image_data_generator.py
```diff
@@ -654,15 +654,18 @@ def flow_from_dataframe(self,
             drop_duplicates=drop_duplicates
         )
 
-    def standardize(self, x):
+    def standardize(self, x, inplace=True):
         """Applies the normalization configuration to a batch of inputs.
 
         # Arguments
             x: Batch of inputs to be normalized.
+            inplace: If to change x in place when doing transformations.
```
gabrieldemarmiesse: Rather than doing this, I think we should modify the docstring. See https://github.com/keras-team/keras/pull/10153 which is a very similar PR. The discussion is roughly the same.

gabrieldemarmiesse: Yes. I believe the problem is that the documentation is not explicit about it. We should make sure this behavior is documented (and maybe explain why we choose to do the modification in-place)  and also provide the workaround `standardize(np.copy(x))`.

Sometimes a good documentation is all we need.

rragundez: done @gabrieldemarmiesse. Let me know if you agree, and I will close the related issue as well.

gglanzani:
@rragundez wasn't there some other piece of the library that was assuming that the input was changed in place?

Otherwise LGTM

gabrieldemarmiesse:
LGTM too.

PR #144, opened 2019-01-20
Improve performance of TimeSeriesGenerator by 1.7x-3x

KuzMenachem:
### Summary
Loads data and targets using a list comprehension which is much faster than a for loop & removes the need to pre-create an empty batch.

My own testing shows speedups of 1.7x-6x across all batch sizes (I've tested 4, 16, 64 and 256), multiple input/target dimensions (I've tested 1, 32 and 256 dimensional inputs & targets) and while loading the data from both disk and memory.

### Related Issues
None
### PR Overview


KuzMenachem:
I'm unsure why the test doesn't pass now, any help is appreciated.

PR #149, opened 2019-01-25
array_to_img fixes

### LINKED ISSUE #148: array_to_image errors
When we convert Image in mode 'L' to a numpy array it has dimensions 2 .
Using array_to_image to convert back it to Image , it gives error .

surgan12:
fixes #148 .

### DIFF
--- a/keras_preprocessing/image/utils.py
+++ b/keras_preprocessing/image/utils.py
@@ -224,17 +224,20 @@ def array_to_img(x, data_format='channels_last', scale=True, dtype='float32'):
         ImportError: if PIL is not available.
         ValueError: if invalid `x` or `data_format` is passed.
     """
+    allowed_dimensions = [2, 3]
     if pil_image is None:
         raise ImportError('Could not import PIL.Image. '
                           'The use of `array_to_img` requires PIL.')
     x = np.asarray(x, dtype=dtype)
-    if x.ndim != 3:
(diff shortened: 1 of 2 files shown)


rragundez:
The docstring is pretty clear on what the function is expecting. Also your propose changes will brake on a certain situation because now you have two places that decide where the channels go: the argument `data_format` and where you add the channels in the code. This can be solve but still I think that this should be handle on the user side based on the docstring and clear exception that is thrown. What do you think?

surgan12:
@rragundez Many a times i think the transforms need to be applied in a pipeline .
IMO, it is better to add 2 d array support like the one torchvision provides.
What do you say ?

rragundez:
i think I agree with you but let me think about it a bit more, in the meantime if you can please update the PR as it's still buggy. You are adding channels last if 2D but the the user can still say `data_format='channels_first'`.

rragundez:
```
from keras_preprocessing.image.utils import array_to_img
import numpy as np

img = np.ones((2, 2))
array_to_img(img, data_format='channels_first')
```

surgan12:
@rragundez sorry , i was not on the same page as you !!

rragundez:
I've thought about it and receiving a 2D array brakes the fact that the parameter `data_format` indicates that there is a channel dimension. 
The best would be to inform the user on what to do if he wants

Issue #154, opened 2019-01-30
Add pandas as a dependency


paulgc:
Import fails when pandas is not already installed.

`from keras_preprocessing import image`

results in

`ImportError: No module named pandas.api.types`

pandas should be specified as a dependency in setup.py

ravelab:
saw this error too, tensorflow in one of my docker containers failed, and I won't be able to add pandas as dependency easily

yangsiran:
Similar problem was met when installing and running tensorflow with keras-preprocessing 1.0.6

rragundez:
Hi thanks very much for pointing it out, actually the import seems to be unused, so it should just be removed.

https://github.com/keras-team/keras-preprocessing/blob/281090d8ab7bdfa47408624c31fad9481db51227/keras_preprocessing/image/dataframe_iterator.py#L11

nhc-motosumo:
This (very temporary!) workaround worked for me: Force old version before installing the package that requires _keras_:
`pip3 install Keras-Preprocessing==1.0.5`

DavidNorman:
actually the above issue isn't exactly the same.  but there is an issue with installing and running TF at the moment due to this missing dependency.

Is there an ETA for a fix?

rragundez:
It has been merged to master. Waiting on feedback about a minor release.

kokes:
@rragundez thanks for the fix! Have you pushed it to pypi? It's still [not showing up](https://pypi.org/project/Keras-Preprocessing/#history), not sure if there's an approval process or something.

Dref360:
@fchollet could you push the new version please? 

We are working on a test that would catch this kind of issue. Sorry for the inconvenience.

Dref360:
Update: There is a new version (1.0.8) on pip.

PR #159, opened 2019-02-02
[WIP] Update CI

rragundez:
### Summary
Update CI, basically separate every test component.
- PEP8 (no dependencies installed at all)
- PyTests (install test dependencies), runs for Python 2.6, 3.6 and Keras PyPI and Keras HEAD
- Imports (install package dependencies), runs for Python 2.6, 3.6. Basically imports the package and modules as a check that nothing breaks.

I also change where some of the logic happens. Before there was some logic in `travis.yml` that in my opinion `shuoldn't` be there but in `setup.py`, in this way the user can actually execute the same that travis is doing by `pip install -e .[tests] && pytest tests` or  `pip install -e .[pep8] && flake8`. So much easier to debug without the need to push and create a pull request to reproduce travis execution (up to some complexity).

This should lead to a more robust check and specially avoid the error given during the release of 1.0.6. That error for example will be catch in two places now: 
- PEP8: as an unused import
- Imports: since pandas is not installed on that check.

NOTE: The phase caching the bug in the 1.0.6 should not be the Pytest phase, that actually should work. Since pandas is needed for unit testing.

### Related Issues
#154 

### PR Overview

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,20 +3,30 @@ dist: trusty
 language: python
 matrix:
     include:
+        # check code style and Python 2.7
         - python: 2.7
           env: TEST_MODE=PEP8
+        # run tests with keras from source and Python 2.7
         - python: 2.7
           env: KERAS_HEAD=true
+          env: TEST_MODE=TESTS
+        # run tests with keras from source and Python 3.6
         - python: 3.6
           env: KERAS_HEAD=true
+          env: TEST_MODE=TESTS
+        # run tests with keras from PyPI and Python 2.7
         - python: 2.7
+          env: TEST_MODE=TESTS
+        # run tests with keras from PyPI and Python 3.6
         - python: 3.6
-install:
-  # code below is taken from http://conda.pydata.org/docs/travis.html
+          env: TEST_MODE=TESTS
+
+before_install:
+  - sudo apt-get update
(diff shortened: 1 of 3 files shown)


rragundez:
In the last commit I added a pandas import to test.

This is the travis output before that import (all checks pass)

https://travis-ci.org/keras-team/keras-preprocessing/builds/487899266?utm_source=github_status&utm_medium=notification

This is the travis output after adding that import

https://travis-ci.org/keras-team/keras-preprocessing/builds/487900770?utm_source=github_status&utm_medium=notification

In this case the combination of PEP8 and IMPORTS testing works great, in one hand if you import something and it fails and it's unused you won't added as a dependency because PEP8 tell you that you don't have to do that since it's unused. On the other hand if it is used but no current dependency then the IMPORTS check will tell you that.

I think this testing framework is pretty nice now.

Dref360:
LGTM, please rollback the previous commit and I'll merge.

rragundez:
@Dref360 done

rragundez:
BTW if you would like me to take a look and do something similar with the other repos let me know.

Dref360:
I guess doing the same for keras would be valuable. Be sure that a clean install is fine. keras-app does not require this kind of testing imo.

Issue #147, opened 2019-01-24
Provide Sample Weights to ImageDataGenerator.flow_from_dataframe


docmarionum1:
I'd like to specify sample weights when using `ImageDataGenerator.flow_from_dataframe`. I see that `flow` supports it, so I could load my images into a numpy array first and use flow. But it seems like it makes sense to allow `flow_from_dataframe` to support it as well.

To implement, `flow_from_dataframe` would optionally take a `weight_col` parameter, and then if `weight_col` is not None, `_get_batches_of_transformed_samples` would return a third item,  the sample weights.

I'd be happy to send a pull request if this is something that y'all agree makes sense.

rragundez:
I was thinking on adding that so this would be a great addition. Please name the column accordingly to the parameter in flow. I think `sample_weights_col`.

rragundez:
@docmarionum1 I created PR #166 that implements this. Please have a look.

docmarionum1:
@rragundez Thanks! Testing it out and it works great!

lluissalord:
Could it be the case that for multi-output it is not working? Because I tried to pass a list or a dictionary to weight_col and it gave me the following error:

> /opt/conda/lib/python3.6/site-packages/keras_preprocessing/image/image_data_generator.py in flow_from_dataframe(self, dataframe, directory, x_col, y_col, weight_col, target_size, color_mode, classes, class_mode, batch_size, shuffle, seed, save_to_dir, save_prefix, save_format, subset, interpolation, validate_filenames, **kwargs)
>     681             subset=subset,
>     682             interpolation=interpolation,
> --> 683             validate_filenames=validate_filenames
>     684         )
>     685 
> 
> /opt/conda/lib/python3.6/site-packages/keras_preprocessing/image/dataframe_iterator.py in __init__(self, dataframe, directory, image_data_generator, x_col, y_col, weight_col, target_size, color_mode, classes, class_mode, batch_size, shuffle, seed, data_format, save_to_dir, save_prefix, save_format, subset, interpolation, dtype, validate_filenames)
>     127         self.dtype = dtype
>     128         # check that inputs match the required class_mode
> --> 129         self._check_params(df, x_col, y_col, weight_col, classes)
>     130         if validate_filenames:  # check which image files are valid and keep them
>     131             df = self._filter_valid_filepaths(df, x_col)
> 
[...]

PR #166, opened 2019-02-07
Add sample weights to DataFrameIterator

rragundez:
### Summary
Currently there is no implementation of sample weights in DataFrameIterator or flow_from_dataframe. This PR adds that functionality. 
For example with this PR is now possible to train a model using sample weights, given that a column of weights is added to the pandas dataframe.
```
generator = ImageDataGenerator().flow_from_dataframe(
    df,
    weight_col='weights'
)
model.fit_generator(generator, epochs=2, steps_per_epoch=10)
```
### Related Issues
Solves #147 

### PR Overview

### DIFF
--- a/keras_preprocessing/image/directory_iterator.py
+++ b/keras_preprocessing/image/directory_iterator.py
@@ -147,3 +147,8 @@ class DirectoryIterator(BatchFromFilesMixin, Iterator):
     @property
     def labels(self):
         return self.classes
+
+    @property  # mixin needs this property to work
+    def sample_weight(self):
+        # no sample weights will be returned
+        return None
--- a/keras_preprocessing/image/dataframe_iterator.py
+++ b/keras_preprocessing/image/dataframe_iterator.py
@@ -169,8 +169,8 @@ class DataFrameIterator(BatchFromFilesMixin, Iterator):
                              .format(self.class_mode, self.allowed_class_modes))
         # check that filenames/filepaths column values are all strings
         if not all(df[x_col].apply(lambda x: isinstance(x, str))):
-            raise ValueError('All values in column x_col={} must be strings.'
-                             .format(x_col))
+            raise TypeError('All values in column x_col={} must be strings.'
+                            .format(x_col))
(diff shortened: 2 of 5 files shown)


keras_preprocessing/image/directory_iterator.py
```diff
@@ -147,3 +147,8 @@ def filepaths(self):
     @property
     def labels(self):
         return self.classes
+
+    @property  # mixin needs this property to work
+    def sample_weight(self):
+        # no sample weights will be returned
+        return None
```
Dref360: Suggestion: Wouldn't returning an array of ones be cleaner?

rragundez: you might be right. Let me take a look how `NumpyArrayIterator` does it, but I think it does something similar, as in not returning any weights at all, hence saving some computation when calculating the loss function down the line.

hbarovertwo:
@rragundez How can this work when doing multi-label classification? Many possible labels and each image can have at least one label. Elements in my labels column in dataframe look like these tuples (lbl1, lbl2, lbl3,...) so corresponding weight elements should look like (weight_lbl1, weight_lbl2,...)?

rragundez:
@hbarovertwo unfortunately you cannot, hence the name sample weights. In this blog I make a similar example https://blog.godatadriven.com/rod-keras-multi-label
basically what I state about class imbalances in multi-label settings

"So what can we do about it?... pray. I won't go into detail but one way to counter the imbalances is with a combination of class weights and sample weights."

Trying to be respectful and honest, I'm also almost sure that that your weights (weight_lbl1, weight_lbl2,...) don't do what you think they do. In order to have balance it has to be across observations and labels, which sometimes has no solution (making one balance better makes the other worst most of the time). The best approach depends on the specifics but I would make sample weights and class weights that give me a good approximation to both balance. Good luck! let me know if you found a nice way to do it.

hbarovertwo:
I was hoping to calculate batch weights on the fly but it didn't work out for me. I may revisit the problem later.

PR #167, opened 2019-02-08
Add numbers and - to image names found by list_pictures.

jnphilipp:
### Summary
Currently list_pictures  only finds images with letters expanding this to include numbers and -.
### PR Overview

### DIFF
--- a/keras_preprocessing/image/utils.py
+++ b/keras_preprocessing/image/utils.py
@@ -137,7 +137,7 @@ def list_pictures(directory, ext=('jpg', 'jpeg', 'bmp', 'png', 'ppm')):
         a list of paths
     """
     if not isinstance(ext, tuple):
-        ext = tuple(ext)
+        ext = (ext,)
     ext = tuple('.%s' % e for e in ext)
     return [os.path.join(root, f)
             for root, _, files in os.walk(directory) for f in files
--- a/keras_preprocessing/image/utils.py
+++ b/keras_preprocessing/image/utils.py
@@ -5,7 +5,6 @@ from __future__ import division
 from __future__ import print_function
 
 import os
-import re
 import warnings
 
 import numpy as np
--- a/keras_preprocessing/image/utils.py
+++ b/keras_preprocessing/image/utils.py
@@ -138,7 +138,7 @@ def list_pictures(directory, ext=('jpg', 'jpeg', 'bmp', 'png', 'ppm')):
     """
     if not isinstance(ext, tuple):
         ext = tuple(ext)
-    ext = tuple(f'.{e}' for e in ext)
+    ext = tuple('.%s' % e for e in ext)
     return [os.path.join(root, f)
             for root, _, files in os.walk(directory) for f in files
             if f.lower().endswith(ext)]
--- a/keras_preprocessing/image/utils.py
+++ b/keras_preprocessing/image/utils.py
@@ -130,7 +130,7 @@ def load_img(path, grayscale=False, color_mode='rgb', target_size=None,
 def list_pictures(directory, ext='jpg|jpeg|bmp|png|ppm'):
     return [os.path.join(root, f)
             for root, _, files in os.walk(directory) for f in files
-            if re.match(r'([\w]+\.(?:' + ext + '))', f.lower())]
+            if re.match(r'([\w\d-]+\.(?:' + ext + '))', f.lower())]
 
 
 def _iter_valid_files(directory, white_list_formats, follow_links):
--- a/keras_preprocessing/image/utils.py
+++ b/keras_preprocessing/image/utils.py
@@ -127,18 +127,21 @@ def load_img(path, grayscale=False, color_mode='rgb', target_size=None,
     return img
 
 
-def list_pictures(directory, ext='jpg|jpeg|bmp|png|ppm'):
+def list_pictures(directory, ext=('jpg', 'jpeg', 'bmp', 'png', 'ppm')):
     """Lists all pictures in a directory, including all subdirectories.
 
     # Arguments
-        directory: Absolute path to the directory
-        ext: extensions of the pictures, separated by '|'
+        directory: string, absolute path to the directory
+        ext: tuple of strings, extensions of the pictures, separated by '|'
     # Returns
         a list of paths
     """
+    if not isinstance(ext, tuple):
+        ext = tuple(ext)
+    ext = tuple(f'.{e}' for e in ext)
     return [os.path.join(root, f)
             for root, _, files in os.walk(directory) for f in files
-            if re.match(r'.*(' + ext + ')$', f.lower())]
+            if f.lower().endswith(ext)]
 
 
 def _iter_valid_files(directory, white_list_formats, follow_links):
--- a/keras_preprocessing/image/utils.py
+++ b/keras_preprocessing/image/utils.py
@@ -127,7 +127,8 @@ def load_img(path, grayscale=False, color_mode='rgb', target_size=None,
     return img
 
 
-def list_pictures(directory, ext=('jpg', 'jpeg', 'bmp', 'png', 'ppm')):
+def list_pictures(directory, ext=('jpg', 'jpeg', 'bmp', 'png', 'ppm', 'tif',
+                                  'tiff')):
     """Lists all pictures in a directory, including all subdirectories.
 
     # Arguments
--- a/keras_preprocessing/image/utils.py
+++ b/keras_preprocessing/image/utils.py
@@ -132,13 +132,11 @@ def list_pictures(directory, ext=('jpg', 'jpeg', 'bmp', 'png', 'ppm')):
 
     # Arguments
         directory: string, absolute path to the directory
-        ext: tuple of strings, extensions of the pictures, separated by '|'
+        ext: tuple of strings or single string, extensions of the pictures
     # Returns
         a list of paths
     """
-    if not isinstance(ext, tuple):
-        ext = (ext,)
-    ext = tuple('.%s' % e for e in ext)
+    ext = tuple('.%s' % e for e in ((ext,) if isinstance(ext, str) else ext))
     return [os.path.join(root, f)
             for root, _, files in os.walk(directory) for f in files
             if f.lower().endswith(ext)]
--- a/keras_preprocessing/image/utils.py
+++ b/keras_preprocessing/image/utils.py
@@ -128,9 +128,17 @@ def load_img(path, grayscale=False, color_mode='rgb', target_size=None,
 
 
 def list_pictures(directory, ext='jpg|jpeg|bmp|png|ppm'):
+    """Lists all pictures in a directory, including all subdirectories.
+
+    # Arguments
+        directory: Absolute path to the directory
+        ext: extensions of the pictures, separated by '|'
+    # Returns
+        a list of paths
+    """
     return [os.path.join(root, f)
             for root, _, files in os.walk(directory) for f in files
-            if re.match(r'([\w\d-]+\.(?:' + ext + '))', f.lower())]
+            if re.match(r'.*(' + ext + ')$', f.lower())]
 
 
 def _iter_valid_files(directory, white_list_formats, follow_links):
--- a/tests/image_test.py
+++ b/tests/image_test.py
@@ -1232,7 +1232,7 @@ class TestImage(object):
         filenames = ['test.png', 'test0.jpg', 'test-1.jpeg', '2test.bmp',
                      '2-test.ppm', '3.png', '1.jpeg', 'test.bmp', 'test0.ppm']
         subdirs = ['', 'subdir1', 'subdir2']
-        images = [os.path.join(tmpdir, subdir, f) for subdir in subdirs
+        images = [tmpdir.mkdir(subdir).join(f) for subdir in subdirs
                   for f in filenames]
 
         for img in images:
--- a/tests/image_test.py
+++ b/tests/image_test.py
@@ -6,6 +6,7 @@ import tempfile
 import shutil
 import pandas as pd
 import random
+import string
 
 from keras_preprocessing import image
 
@@ -1224,6 +1225,21 @@ class TestImage(object):
             loaded_im = image.load_img(filename_rgb, target_size=(25, 25),
                                        interpolation="unsupported")
 
+
+    def test_list_pictures(self, tmpdir):
+        vocab = list(string.ascii_lowercase + string.ascii_uppercase +
+                     string.digits + '-')
+        images = []
+
+        for i in range(100):
+            images.append(os.path.join(tmpdir,
+                                       f'{"".join(np.random.choice(vocab,
+                                                                   10))}.png'))
+            image.array_to_img(np.array(255 * np.random.rand(100, 100, 3),
+                               dtype=np.uint8), scale=False).save(images[-1])
+
+        found_images = image.list_pictures(tmpdir)
+        assert len(found_images) == len(images)
 
 if __name__ == '__main__':
     pytest.main([__file__])
--- a/tests/image_test.py
+++ b/tests/image_test.py
@@ -1227,16 +1227,15 @@ class TestImage(object):
 
 
     def test_list_pictures(self, tmpdir):
-        vocab = list(string.ascii_lowercase + string.ascii_uppercase +
-                     string.digits + '-')
-        images = []
+        images.append(os.path.join(tmpdir, 'test.png'))
+        images.append(os.path.join(tmpdir, 'test0.jpg'))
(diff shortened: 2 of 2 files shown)


rragundez:
Can you write a test for this change? thanks

keras_preprocessing/image/utils.py
```diff
@@ -130,7 +130,7 @@ def load_img(path, grayscale=False, color_mode='rgb', target_size=None,
 def list_pictures(directory, ext='jpg|jpeg|bmp|png|ppm'):
     return [os.path.join(root, f)
             for root, _, files in os.walk(directory) for f in files
-            if re.match(r'([\w]+\.(?:' + ext + '))', f.lower())]
+            if re.match(r'([\w\d-]+\.(?:' + ext + '))', f.lower())]
```
rragundez: The problem is not in the regex but in the use of `match`, as it only matches at the beginning. Also the main use of the function is the check of the extensions. There are easier and clearer ways to do this, like `.endswith`.

keras_preprocessing/image/utils.py
```diff
@@ -130,7 +130,7 @@ def load_img(path, grayscale=False, color_mode='rgb', target_size=None,
 def list_pictures(directory, ext='jpg|jpeg|bmp|png|ppm'):
```
rragundez: Docstring is missing

tests/image_test.py
```diff
@@ -1225,5 +1226,20 @@ def test_load_img(self, tmpdir):
                                        interpolation="unsupported")
 
 
+    def test_list_pictures(self, tmpdir):
```
rragundez: This test is not idempotent. Is very unlikely but it could be that the test passes. Is better to define some cases that we need to check and use those ones instead of a random generation of filenames.

If this is a test for `list_pictures` is better to not use other functions in the test like `array_to_img` and `save`.

rragundez:
let me know if you will implement the changes, if not I will make the changes. thanks

jnphilipp:
I'll do the changes.

tests/image_test.py
jnphilipp: The ```save``` is from ```Pillow``` not ```utils```.

What where you thinking of and how many?
```test.png```, ```test0.png```, ```test-1.png```, ```1test.jpg```, ```1-test.jpg``` enough?

Comment (rragundez) [reply tests/image_test.py]: `-` at the end? I think does look good.

File: keras_preprocessing/image/utils.py [hunk 4]
```diff
@@ -128,9 +128,17 @@ def load_img(path, grayscale=False, color_mode='rgb', target_size=None,
 
 
 def list_pictures(directory, ext='jpg|jpeg|bmp|png|ppm'):
+    """Lists all pictures in a directory, including all subdirectories.
+
+    # Arguments
+        directory: Absolute path to the directory
```
rragundez: please start with the type of the argument

tests/image_test.py
```diff
@@ -1227,16 +1227,15 @@ def test_load_img(self, tmpdir):
...
-                                       f'{"".join(np.random.choice(vocab,
-                                                                   10))}.png'))
-            image.array_to_img(np.array(255 * np.random.rand(100, 100, 3),
-                               dtype=np.uint8), scale=False).save(images[-1])
+        images.append(os.path.join(tmpdir, 'test.png'))
+        images.append(os.path.join(tmpdir, 'test0.jpg'))
+        images.append(os.path.join(tmpdir, 'test-1.jpeg'))
+        images.append(os.path.join(tmpdir, '2test.bmp'))
+        images.append(os.path.join(tmpdir, '2-test.ppm'))
+        images.append(os.path.join(tmpdir, '3.png'))
```
rragundez: `images = [os.path.join(tmpdir, filename) for filename in filenames]`

tests/image_test.py
```diff
@@ -1227,16 +1227,15 @@ def test_load_img(self, tmpdir):
...
+
+        for img in images:
```
rragundez: `for img in (os.path.join(tmpdir, f) for f in filenames)`

tests/image_test.py
```diff
@@ -1227,16 +1227,15 @@ def test_load_img(self, tmpdir):
...
+            Image.new('RGB', (32, 32)).save(img)
```
rragundez: I don't think we need to actually save images right? they can be empty files.

keras_preprocessing/image/utils.py
```diff
@@ -128,9 +128,17 @@ def load_img(path, grayscale=False, color_mode='rgb', target_size=None,
...
+        ext: extensions of the pictures, separated by '|'
+    # Returns
+        a list of paths
+    """
     return [os.path.join(root, f)
             for root, _, files in os.walk(directory) for f in files
-            if re.match(r'([\w\d-]+\.(?:' + ext + '))', f.lower())]
+            if re.match(r'.*(' + ext + ')$', f.lower())]
```
rragundez: I think it is clearer to do `f.endswith(tuple_of_extensions_with_dot)`, what do you think? perhaps is also faster

jnphilipp: Don't know if its faster, but clearer yeah. Should I also change the function parameter to be a tuple or should I split the string.

tests/image_test.py
jnphilipp: Should work.

rragundez:
Tuple is fine because it's immutable. Without the dot. Let's add the dot on the endswith logic. Please also check that it can receive a single extension from the user.

keras_preprocessing/image/utils.py
rragundez: Tuple is fine because it's immutable. Without the dot. Let's add the dot on the endswith logic. Please also check that it can receive a single extension from the user.

keras_preprocessing/image/utils.py
Vijayabhaskar96: what about tiff?

keras_preprocessing/image/utils.py
rragundez: seems to be faster

selection_047

keras_preprocessing/image/utils.py
```diff
@@ -127,18 +127,21 @@ def load_img(path, grayscale=False, color_mode='rgb', target_size=None,
...
 
-def list_pictures(directory, ext='jpg|jpeg|bmp|png|ppm'):
+def list_pictures(directory, ext=('jpg', 'jpeg', 'bmp', 'png', 'ppm')):
     """Lists all pictures in a directory, including all subdirectories.
 
     # Arguments
-        directory: Absolute path to the directory
-        ext: extensions of the pictures, separated by '|'
+        directory: string, absolute path to the directory
+        ext: tuple of strings, extensions of the pictures, separated by '|'
```
rragundez: no longer separated by `|`

tests/image_test.py
Vijayabhaskar96: why 'tif' and 'tiff' are missing here and in the function `list_pictures` ? Is this a mistake or am I missing something?

keras_preprocessing/image/utils.py
```diff
@@ -127,18 +127,21 @@ def load_img(path, grayscale=False, color_mode='rgb', target_size=None,
...
     # Returns
         a list of paths
     """
+    if not isinstance(ext, tuple):
+        ext = tuple(ext)
+    ext = tuple(f'.{e}' for e in ext)
```
rragundez: f-strings do not work with python 2.7

keras_preprocessing/image/utils.py
```diff
@@ -127,10 +127,21 @@ def load_img(path, grayscale=False, color_mode='rgb', target_size=None,
     return img
 
 
-def list_pictures(directory, ext='jpg|jpeg|bmp|png|ppm'):
+def list_pictures(directory, ext=('jpg', 'jpeg', 'bmp', 'png', 'ppm')):
```
Vijayabhaskar96: tif and tiff?

tests/image_test.py
rragundez: it has been this way since the beginning https://github.com/keras-team/keras-preprocessing/blame/014457b52def2af30bff88809437f8941451b76f/keras_preprocessing/image.py#L542
and I'm not sure we should change it since there is no issue related as we can break some systems with the default behavior. What do you think?

keras_preprocessing/image/utils.py
rragundez: it has been this way since the beginning https://github.com/keras-team/keras-preprocessing/blame/014457b52def2af30bff88809437f8941451b76f/keras_preprocessing/image.py#L542
and I'm not sure we should change it since there is no issue related as we can break some systems with the default behavior. What do you think?

rragundez: it has been this way since the beginning https://github.com/keras-team/keras-preprocessing/blame/014457b52def2af30bff88809437f8941451b76f/keras_preprocessing/image.py#L542
and I'm not sure we should change it since there is no issue related as we can break some systems with the default behavior. What do you think?

tests/image_test.py
```diff
@@ -1225,5 +1226,29 @@ def test_load_img(self, tmpdir):
...
+        images.append(os.path.join(tmpdir, 'class1', 'test0.jpg'))
+        images.append(os.path.join(tmpdir, 'class1', '1.jpeg'))
+
+        os.mkdir(os.path.join(tmpdir, 'class2'))
+        images.append(os.path.join(tmpdir, 'class2', 'test.bmp'))
+        images.append(os.path.join(tmpdir, 'class2', 'test0.ppm'))
+        images.append(os.path.join(tmpdir, 'class2', '1.jpeg'))
+
+        for img in images:
+            os.mknod(img)
```
rragundez: this only works with unix systems

tests/image_test.py
```diff
@@ -1225,5 +1226,29 @@ def test_load_img(self, tmpdir):
                                        interpolation="unsupported")
 
 
+    def test_list_pictures(self, tmpdir):
+        images = [os.path.join(tmpdir, 'test.png')]
```
rragundez: you can create all the images in a single list comprehension as I indicated in my other comment

tests/image_test.py
```diff
@@ -1225,5 +1226,29 @@ def test_load_img(self, tmpdir):
...
+
+        os.mkdir(os.path.join(tmpdir, 'class2'))
+        images.append(os.path.join(tmpdir, 'class2', 'test.bmp'))
+        images.append(os.path.join(tmpdir, 'class2', 'test0.ppm'))
+        images.append(os.path.join(tmpdir, 'class2', '1.jpeg'))
+
+        for img in images:
+            os.mknod(img)
+
+        found_images = image.list_pictures(tmpdir)
```
rragundez: please also test with a single extension

keras_preprocessing/image/utils.py
Vijayabhaskar96: Iterator accepts tif and tiff images, I don't see any reason why it should be left out here and also list_pictures function is not used in any internal code, so changing it doesn't going to break anything I believe.

rragundez: I understand and agree with your point of being used in Iterator but I meant non-internal code but code of others. 
But sure let's add it. @jnphilipp

jnphilipp: Has anyone a tiff image to test if ``load``` and ```img_to_array``` work. If so I would add it, if not then not.

tests/image_test.py
```diff
@@ -1225,5 +1226,29 @@ def test_load_img(self, tmpdir):
...
+
+        filenames = ['test.png', 'test0.jpg', 'test-1.jpeg', '2test.bmp',
+                     '2-test.ppm', '3.png']
+        images = [os.path.join(tmpdir, f) for f in filenames]
+
+        class1 = ['test.png', 'test0.jpg', '1.jpeg']
+        images += [os.path.join(tmpdir, 'class1', f) for f in class1]
+
+        class2 = ['test.bmp', 'test0.ppm', '1.jpeg']
+        images += [os.path.join(tmpdir, 'class2', f) for f in class2]
```
rragundez: no need for all that boiler code. `images = [os.path.join(tmpdir, subdir, f) for subdir in ['', 'subdir1', 'subdir2'] for f in filenames]`

tests/image_test.py
```diff
@@ -1224,6 +1225,26 @@ def test_load_img(self, tmpdir):
...
                                        interpolation="unsupported")
 
+    def test_list_pictures(self, tmpdir):
+        os.mkdir(os.path.join(tmpdir, 'class1'))
+        os.mkdir(os.path.join(tmpdir, 'class2'))
+
+        filenames = ['test.png', 'test0.jpg', 'test-1.jpeg', '2test.bmp',
+                     '2-test.ppm', '3.png', '1.jpeg', 'test.bmp', 'test0.ppm']
+        subdirs = ['', 'subdir1', 'subdir2']
+        images = [tmpdir.mkdir(subdir).join(f) for subdir in subdirs
```
rragundez: This won't work. `tmpdir` will raise an exception if trying to create an existing dirextory. Your test can be this:
```
        filenames = ['test.png', 'test0.jpg', 'test-1.jpeg',
                     '2test.bmp', '2-test.ppm', '3.png']
        subdirs = [tmpdir.mkdir(s) for s in ['subdir1', 'subdir2']] + [tmpdir]
        images = [d.join(f).write('') for d in subdirs for f in filenames]

        found_images = image.list_pictures(tmpdir)
        assert len(found_images) == len(images)

        found_images = image.list_pictures(tmpdir, ext='png')
        assert len(found_images) == 6
```

keras_preprocessing/image/utils.py
```diff
@@ -127,10 +127,21 @@ def load_img(path, grayscale=False, color_mode='rgb', target_size=None,
...
+
+    # Arguments
+        directory: string, absolute path to the directory
+        ext: tuple of strings, extensions of the pictures, separated by '|'
+    # Returns
+        a list of paths
+    """
+    if not isinstance(ext, tuple):
+        ext = (ext,)
+    ext = tuple('.%s' % e for e in ext)
```
rragundez: better to check if string
`ext = tuple('.%s' % e for e in ((ext,) if isinstance(ext, str) else ext))`

keras_preprocessing/image/utils.py
```diff
@@ -127,10 +127,21 @@ def load_img(path, grayscale=False, color_mode='rgb', target_size=None,
     return img
 
 
-def list_pictures(directory, ext='jpg|jpeg|bmp|png|ppm'):
+def list_pictures(directory, ext=('jpg', 'jpeg', 'bmp', 'png', 'ppm')):
+    """Lists all pictures in a directory, including all subdirectories.
+
+    # Arguments
+        directory: string, absolute path to the directory
+        ext: tuple of strings, extensions of the pictures, separated by '|'
```
rragundez: Update docstring to remove the pipe operator in the descriptions and to say that it can also be a single string with an extension.

keras_preprocessing/image/utils.py
rragundez: please add the `tif` and `tiff` extension

tests/image_test.py
```diff
@@ -1224,6 +1225,23 @@ def test_load_img(self, tmpdir):
...
+    def test_list_pictures(self, tmpdir):
+        filenames = ['test.png', 'test0.jpg', 'test-1.jpeg', '2test.bmp',
+                     '2-test.ppm', '3.png', '1.jpeg', 'test.bmp', 'test0.ppm']
+        subdirs = ['', 'subdir1', 'subdir2']
+        images = [tmpdir.ensure(subdir, dir=True).join(f) for subdir in subdirs
+                  for f in filenames]
+
+        for img in images:
+            with open(img, 'w') as f:
+                f.write('\n')
```
rragundez: No need for all these:
```
 subdirs = ['', 'subdir1', 'subdir2']
images = [tmpdir.ensure(subdir, f) for subdir in subdirs for f in filenames]
```

tests/image_test.py
```diff
@@ -1224,6 +1225,23 @@ def test_load_img(self, tmpdir):
             loaded_im = image.load_img(filename_rgb, target_size=(25, 25),
                                        interpolation="unsupported")
 
+    def test_list_pictures(self, tmpdir):
+        filenames = ['test.png', 'test0.jpg', 'test-1.jpeg', '2test.bmp',
+                     '2-test.ppm', '3.png', '1.jpeg', 'test.bmp', 'test0.ppm']
```
rragundez: add `.tif` and `.tiff` filenames, and a couple of filenames that should not be retrieved by the function, than change the assertions accordingly.

rragundez:
LGTM, thanks!

PR #168, opened 2019-02-08
Add multi-output support to `flow_from_dataframe`

rragundez:
### Summary

At the moment `flow_from_dataframe` does not support multi-output, i think is the excellent candidate to do that. Basically if the user sets the mode as `multi_output` then he needs to pass a list of columns and each column will represent each output. This PR enables that functionality. It is now possible to seemingly train multi output. 

In the self-contained example below I show how you could train:
- Two outputs which are single number
- Two outputs, one a single number and another a 1D array
- Two outputs, one a single number and another a 2D array

This functionality will enable several use cases (related to multi-task learning) which currently are not possible or they need to be done in a hacky way.

```
import os
import random
import numpy as np
import pandas as pd
from keras.models import Model
from keras.layers import Flatten, Dense, Input, Reshape

filenames = []
for i in range(20):
    filename = '/tmp/{}.jpg'.format(i)
    plt.imsave(filename, np.random.uniform(size=(3, 3, 3)))
    filenames.append(filename)

classes = ['dog', 'cat'] * 10
df = pd.DataFrame({'filename': filenames, 'class': classes}).sample(frac=1).reset_index(drop=True)
print(df.dtypes)
df.head()

output = '2d_array' # if mix then one ouput is a scalar the other an image
if output == 'real':
    df = df.assign(output_0=np.random.uniform(size=len(df)), output_1=np.random.uniform(size=len(df)))
elif output == '1d_array':
    df = df.assign(output_0=np.random.uniform(size=len(df)), output_1=[np.random.uniform(size=(2, 2, 1)).flatten() for _ in range(len(df))])
elif output == '2d_array':
    df = df.assign(output_0=np.random.uniform(size=len(df)), output_1=[np.random.uniform(size=(2, 2, 1)) for _ in range(len(df))])

batch_size = 2
generator = ImageDataGenerator(rescale=1/255).flow_from_dataframe(
    df,
    y_col=['output_0', 'output_1'],
    batch_size=batch_size,
    target_size=(3, 3),
    shuffle = False,
    class_mode='multi_output',
)
batch = next(generator)
assert len(batch) == 2
assert np.array_equal(batch[1][0], np.array(df['output_0'].tolist()[:batch_size]))
assert np.array_equal(batch[1][1], np.array(df['output_1'].tolist()[:batch_size]))

inputs = Input(shape=(3, 3 ,3))
x = Flatten()(inputs)
prediction_0 = Dense(units=1, name="prediction_0")(x)
if output == 'real':
    prediction_1 = Dense(units=1, name="prediction_1")(x)
elif output == '1d_array':
    prediction_1 = Dense(units=4, name="prediction_1")(x)
elif output == '2d_array':
    x = Dense(units=4)(x)
    prediction_1 = Reshape((2, 2, 1), name="prediction_1")(x)

model = Model(inputs=inputs, outputs=[prediction_0, prediction_1])

losses = {
    "prediction_0": "mean_squared_error",
    "prediction_1": "binary_crossentropy",
}
loss_weights = {"prediction_0": 1.0, "prediction_1": 1.0}
model.compile(optimizer='adam', loss=losses, loss_weights=loss_weights, metrics=["accuracy"])

model.fit_generator(generator, epochs=2, steps_per_epoch=10)
model.predict_generator(generator, steps=1)
```
### Related Issues
Solves #99 

### PR Overview

- [y ] This PR requires to update the documentation [y/n] (make sure the docs are up-to-date)
- [n ] This PR changes the current API [y/n] (all API changes need to be approved by fchollet)

### DIFF
--- a/keras_preprocessing/image/dataframe_iterator.py
+++ b/keras_preprocessing/image/dataframe_iterator.py
@@ -46,8 +46,8 @@ class DataFrameIterator(BatchFromFilesMixin, Iterator):
             Color mode to read images.
         classes: Optional list of strings, classes to use (e.g. `["dogs", "cats"]`).
             If None, all classes in `y_col` will be used.
-        class_mode: one of "categorical", "binary", "sparse", "input",
-            "other" or None. Default: "categorical".
+        class_mode: one of "binary", "categorical", "input", "multi_output",
+            "raw", sparse" or None. Default: "categorical".
             Mode for yielding the targets:
             - `"binary"`: 1D numpy array of binary labels,
             - `"categorical"`: 2D numpy array of one-hot encoded labels.
--- a/keras_preprocessing/image/dataframe_iterator.py
+++ b/keras_preprocessing/image/dataframe_iterator.py
@@ -174,7 +174,8 @@ class DataFrameIterator(BatchFromFilesMixin, Iterator):
         # check that y_col has several column names if class_mode is multi_output
         if (self.class_mode == 'multi_output') and not isinstance(y_col, list):
             raise TypeError(
-                'If class_mode="{}", y_col must be a list. Received {}.'
+                'If class_mode="{}", y_col must be a list. Received {}. '
+                'For regression use class_mode="raw".'
                 .format(self.class_mode, type(y_col).__name__)
             )
         # check that filenames/filepaths column values are all strings
@@ -205,8 +206,8 @@ class DataFrameIterator(BatchFromFilesMixin, Iterator):
                 raise TypeError('If class_mode="{}", y_col="{}" column '
                                 'values must be type string, list or tuple.'
                                 .format(self.class_mode, y_col))
-        # raise warning if classes are given and class_mode other or input
-        if classes and self.class_mode in {"input", None}:
+        # raise warning if classes are given but will be unused
+        if classes and self.class_mode in {"input", "multi_output", "raw", None}:
             warnings.warn('`classes` will be ignored given the class_mode="{}"'
                           .format(self.class_mode))
         # check that if weight column that the values are numerical
--- a/keras_preprocessing/image/dataframe_iterator.py
+++ b/keras_preprocessing/image/dataframe_iterator.py
@@ -174,8 +174,7 @@ class DataFrameIterator(BatchFromFilesMixin, Iterator):
         # check that y_col has several column names if class_mode is multi_output
         if (self.class_mode == 'multi_output') and not isinstance(y_col, list):
             raise TypeError(
-                'If class_mode="{}", y_col must be a list. Received {}. '
-                'For regression use class_mode="raw".'
+                'If class_mode="{}", y_col must be a list. Received {}.'
                 .format(self.class_mode, type(y_col).__name__)
             )
         # check that filenames/filepaths column values are all strings
--- a/keras_preprocessing/image/dataframe_iterator.py
+++ b/keras_preprocessing/image/dataframe_iterator.py
@@ -52,10 +52,11 @@ class DataFrameIterator(BatchFromFilesMixin, Iterator):
             - `"binary"`: 1D numpy array of binary labels,
             - `"categorical"`: 2D numpy array of one-hot encoded labels.
                 Supports multi-label output.
-            - `"sparse"`: 1D numpy array of integer labels,
             - `"input"`: images identical to input images (mainly used to
                 work with autoencoders),
             - `"multi_output"`: list with the values of the different columns,
+            - `"raw"`: numpy array of values in `y_col` column(s),
+            - `"sparse"`: 1D numpy array of integer labels,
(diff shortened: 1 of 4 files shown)


adrianodennanni:
Nice job! It was very clever to add a specific loss function for each output.

keras_preprocessing/image/dataframe_iterator.py
```diff
@@ -51,7 +53,7 @@ class DataFrameIterator(BatchFromFilesMixin, Iterator):
             - `"sparse"`: 1D numpy array of integer labels,
             - `"input"`: images identical to input images (mainly used to
                 work with autoencoders),
-            - `"other"`: numpy array of `y_col` data,
```
Dref360: I would like to keep "other", or at least deprecate it.

rragundez: ah yeah, you're right. Let me think about it and propose if to deprecated or keep it, keeping in mind the future application for it. thanks!

Vijayabhaskar96: If 'other' is removed, is there any other way to do regression?

rragundez: I think we should keep it in some way, for regression as you mentioned for example. Let me think how exactly. The `other` word doesn't seem appropriate to me anymore, as it will be the use case for setting a single `y_col` column and returning that without any conversion. So it's very specific `other` I think refers to many types of output, perhaps `raw` instead of `other`? what do you think?

rragundez: done

keras_preprocessing/image/dataframe_iterator.py
```diff
@@ -46,16 +46,17 @@ class DataFrameIterator(BatchFromFilesMixin, Iterator):
             Color mode to read images.
         classes: Optional list of strings, classes to use (e.g. `["dogs", "cats"]`).
             If None, all classes in `y_col` will be used.
-        class_mode: one of "categorical", "binary", "sparse", "input",
-            "other" or None. Default: "categorical".
+        class_mode: one of "binary", "categorical", "input", "multi_output",
+            "raw", sparse" or None. Default: "categorical".
```
Dref360: "sparse"*

tests/image_test.py
```diff
@@ -738,6 +713,114 @@ def test_dataframe_iterator_class_mode_categorical_multi_label(self, tmpdir):
...
+            df, y_col=['output_0', 'output_1'], directory=str(tmpdir),
+            batch_size=3, shuffle=False, class_mode='raw'
+        )
+        batch_x, batch_y = next(df_iterator)
+        assert isinstance(batch_x, np.ndarray)
+        assert len(batch_x.shape) == 4
+        assert isinstance(batch_y, np.ndarray)
+        assert batch_y.shape == (3, 2)
+        print(batch_y)
+        print(df[['output_0', 'output_0']].values[:3])
```
Vijayabhaskar96: Is printing them(819-820) necessary? I don't think why is there a need to print them considering 821 asserts them.

keras_preprocessing/image/dataframe_iterator.py
```diff
@@ -173,6 +171,13 @@ def _check_params(self, df, x_col, y_col, weight_col, classes):
         if self.class_mode not in self.allowed_class_modes:
             raise ValueError('Invalid class_mode: {}; expected one of: {}'
                              .format(self.class_mode, self.allowed_class_modes))
+        # check that y_col has several column names if class_mode is multi_output
+        if (self.class_mode == 'multi_output') and not isinstance(y_col, list):
+            raise TypeError(
+                'If class_mode="{}", y_col must be a list. Received {}. '
+                'For regression use class_mode="raw".'
```
Vijayabhaskar96: I don't think this line is necessary or otherwise could be rephrased better, This statement may confuse people if they're trying to fit a multi-output regression model.

Vijayabhaskar96: I think the statements under `if class_mode not in ["input", "multi_output", "raw", None]:` at Line 147 can be moved under Line 135 which does the same check. Not necessarily to be fixed in this PR though.

rragundez: they are the same checks indeed but unfortunately they cannot be under the same if statement because of the split functionality. As I mentioned before in other PR's the problem here are the naming of variables which is confusing, in the first if statement we collect the classes and in the second we collect the labels (but both have the classes variables name).

Vijayabhaskar96:
@Dref360 Looks good to me. @rragundez Thanks!

viet2411:
Hello,
I'm using flow_from_dataframe with class='multi_output', but when I called the class_indices, it returned:
AttributeError:'DataFrameIterator' object has no attribute 'class_indices'.
How can I solve this problem?

Issue #171, opened 2019-02-15
flow_from_dataframe too slow


qinst64:
I am running the latest version aka __Keras-Preprocessing==1.0.9__ on around 150K images.
the __flow_from_directory__ takes <6s to init whereas __flow_from_dataframe__ takes >30s.

After testing the source code, the line  
```
existence_check = filepaths.map(os.path.isfile)
``` 

is consuming almost all the time. This is a known issue of file check as I  [google it](https://www.google.com/search?q=os.path.isfile%20slow)

After removing the ___filter_valid_filepaths__ (i.e. format_check & existence_check) my code takes < 0.5s.

Any methods to speed up?

If not, a switch for users might be a solution(I am pretty sure all the files in my dataframe  are valid files and  not directories.). thx

Dref360:
So I guess that the way to solve that is to group the paths by their parent folder, call listdir on that folder and compare both list. Would be much faster.

I'll try to submit a PR over the weekend.

jkthompson:
I just ran into the same issue as @qinst64.  In my case I'm working with a directory of over 1 million images on a file system with poor latency, and a directory hierarchy several levels deep.  I'm concerned that listing all files in the directory could also be slow, especially if my dataframe only contains a small fraction of the images in the directory.  In my view, it would be nice to have the option to turn off image validation.  Maybe the generator could catch & skip invalid images at generation time?  Would be happy to help with testing.

rragundez:
I haven't looked at the code specifically but it sounds to me that the best solution will be to:

- use glob.iglob if directory is given. This will be faster and avoid the memory footprint of listing all paths. Also remember to use iglob since we need a generator and glob.glob in python 2.7 returns a list. 
- Keep current behavior if no directory is given. If no directory is given the images could be coming from different directories. 

This difference should be documented to the use stating that it is preferred to give a directory parameter if all images are coming from the same directory.

As a side note we can also restrict the API and force the user to always use a directory parameter, so forcing the user to always have the images in the same directory.

Dref360:
Couldn't just set '/' as the directory and then we need to glob the entire hard drive?

Unfortunatly, I was not able to work on this over the weekend. Will do my best to finish this over the week.

rragundez:
I don't understand the question. What the directory parameters is, is up to the user. I was just mentioning that instead of listdir you do glob in your idea, you only address those subdirecttories mentioned in the filenames column.

jkthompson:
[...]

Issue #172, opened 2019-02-16
Video Generator


brandonschabell:
I have a feature request: 
I'm starting to work with videos more often and I'm spending far too long creating generators that can tie multiple frames together into 4D tensors to be used for video-based RNNs. It would be very helpful to have a video_data_generator. My initial vision of this would rely on a directory structure such as this:

/Train
-- /Class1
-- -- /Video1
-- -- -- /Video1-frame01.jpeg
-- -- -- /Video1-frame02.jpeg
-- -- /Video2
... and so on.

The generator would probably need to include an argument for a maximum sequence length. The generator would then yield 4D np.arrays -> (sequence_length, height, width, channels). I'd be happy to work on the feature, but I wanted to at least see if you felt it was a useful feature or if there are any clever workarounds that I'm not aware of.

- [y] Check that you are up-to-date with the master branch of keras-preprocessing. You can update with:
`pip install git+git://github.com/keras-team/keras-preprocessing.git --upgrade --no-deps`

- [n/a] Provide a link to a GitHub Gist of a Python script that can reproduce your issue (or just copy the script here if it is short).

PR #177, opened 2019-02-22
Add validate filenames parameter to DataFrameIterator

rragundez:
### Summary
This PR allows the user to skip the validation of images filepaths in the pandas DataFrame to speed up the instantiation process. Slow instantiation times have been reported (#171) for a big amount of filenames.  
In addition this PR adds a warning to the user if validation is on and non-valid image filenames are found.
By default image filenames are validated.

### Related Issues
#171 

### PR Overview

### DIFF
--- a/keras_preprocessing/image/dataframe_iterator.py
+++ b/keras_preprocessing/image/dataframe_iterator.py
@@ -80,9 +80,9 @@ class DataFrameIterator(BatchFromFilesMixin, Iterator):
             supported. If PIL version 3.4.0 or newer is installed, "box" and
             "hamming" are also supported. By default, "nearest" is used.
         drop_duplicates: Boolean, whether to drop duplicate rows based on filename.
-        validate_images: Boolean, whether to validate image filenames in `x_col`. If
-        `True`, invalid images will be ignored. Disabling this option can lead to
-        speed-up in the instantiation of this class. Default: `True`.
+        validate_filenames: Boolean, whether to validate image filenames in
+        `x_col`. If `True`, invalid images will be ignored. Disabling this option
+        can lead to speed-up in the instantiation of this class. Default: `True`.
     """
     allowed_class_modes = {
         'categorical', 'binary', 'sparse', 'input', 'other', None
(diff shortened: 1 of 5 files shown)


keras_preprocessing/image/utils.py
```diff
@@ -33,12 +33,17 @@
...
-    There are newer methods to achieve this but this method is backwards compatible.
+    # Arguments
+        filename: String, absolute path to a file
+        white_list_formats: Set, allowed file extensions
+
+    # Returns
+        A boolean value indicating if the filename is valid or not
     """
-    return os.path.splitext(filename)[1].strip('.').lower()
+    return filename.endswith(white_list_formats) and os.path.isfile(filename)
```
jkthompson: want to add .lower() here as in line #172?

keras_preprocessing/image/dataframe_iterator.py
```diff
@@ -80,6 +80,9 @@ class DataFrameIterator(BatchFromFilesMixin, Iterator):
             supported. If PIL version 3.4.0 or newer is installed, "box" and
             "hamming" are also supported. By default, "nearest" is used.
         drop_duplicates: Boolean, whether to drop duplicate rows based on filename.
+        validate_images: Boolean, whether to validate image filenames in `x_col`.
+            Disabling this option can lead to speed-up in the instantiation of
```
jkthompson: Maybe add "Invalid images will be removed." So that this is clear.

rragundez:
Please have another look, I changes the parameter to `validate_filenames`, seemed more appropriate since we are not validating images themselves.

tests/image_test.py
```diff
@@ -590,6 +590,28 @@ def test_valid_args(self):
         with pytest.raises(ValueError):
             image.ImageDataGenerator(brightness_range=0.1)
 
+    def test_dataframe_iterator_validate_validate_filenames(self, tmpdir):
```
jkthompson:
```suggestion
    def test_dataframe_iterator_validate_filenames(self, tmpdir):
```

rragundez: ups, I've made the change.

CA4GitHub:
When `validate_filenames=True` it really can slows= down the call to `flow_from_dataframe`.  For this reason I think `validate_filenames=False` should be the default.  I wonder how many users spend time waiting for this method without realizing how much time can be saved if validating filenames is not ncessary.

Issue #181, opened 2019-03-08
Allow oversampled and/or under-sampled input to flow_from_dataframe


ResidentMario:
I have an image dataset with an unbalanced class distribution: certain common classes have up to 10x as many samples as certain uncommon classes. I would like to rebalance what images my classifier is exposed to by using some combination of oversampling and undersampling methods from `imbalanced-learn`.

I used the following code to see if this was something I could achieve using `flow_from_dataframe`:

```python
X_meta = pd.read_csv("X_meta")  # get source metadata
import numpy as np

split_ratio = 0.8
n_samples = len(X)
split_idx = int(split_ratio * n_samples)

np.random.seed(42)
idxs = np.arange(len(X))
np.random.shuffle(idxs)

X_train, X_test = X[idxs[:split_idx]], X[idxs[split_idx:]]
y_train, y_test = y[idxs[:split_idx]], y[idxs[split_idx:]]

X = X_meta[['CroppedImageURL']].values
y = X_meta['LabelName'].values

from imblearn.over_sampling import RandomOverSampler
X, y = RandomOverSampler(random_state=42).fit_resample(X, y)

import keras
from keras.preprocessing.image import ImageDataGenerator

train_datagen = ImageDataGenerator(
    rotation_range=40,
    width_shift_range=0.2,
    height_shift_range=0.2,
    rescale=1/255,
    shear_range=0.2,
    zoom_range=0.2,
    horizontal_flip=True,
    fill_mode='nearest'
)

test_datagen = ImageDataGenerator(
    rescale=1/255
)

X_train_df = pd.DataFrame().assign(ImagePath=X_train[:, 0], ImageClass=y_train)
X_test_df = pd.DataFrame().assign(ImagePath=X_test[:, 0], ImageClass=y_test)

train_generator = train_datagen.flow_from_dataframe(
    X_train_df,
    directory='/Users/alex/Desktop/comet/data/images_cropped/',
    x_col='ImagePath',
    y_col='ImageClass',
    target_size=(48, 48),
    batch_size=512,
    class_mode='categorical'
)

validation_generator = test_datagen.flow_from_dataframe(
    X_test_df,
    directory='/Users/alex/Desktop/comet/data/images_cropped/',
    x_col='ImagePath',
    y_col='ImageClass',    
    target_size=(48, 48),
    batch_size=512,
    class_mode='categorical'
)
```

[...]

Issue #175, opened 2019-02-22
Is it possible using flow_from_dataframe() to pass the same image-path multiple times to an ImageDataGenerator?


aarondt:
I am trying to balance an unbalanced dataset for image classification. As such, I duplicated the entries of the minority class within my dataframe n-times to match the number of samples of the majority class. This procedure leads to a situation where I have multiple images sharing the same path in my dataframe.

For example my dataframe looks like this:

```
path                          class
C:/image/1.jpg                 "A"
C:/image/1.jpg                 "A
C:/image/2.jpg                 "B"
C:/image/2.jpg                 "B"
```

Now, once I pass the dataframe to my `ImageDataGenerator `using `flow_from_dataframe `it only returns each image exactly once. For example the output would be:

`Found 2 images belonging to 2 classes.`

wheres I would like to have:

`Found 4 images belonging to 2 classes.`

As such, I would like to know if it is possible to use an image multiple times during training using` flow_from_dataframe()` without actually copying the file on hard-drive multiple times? Considering that the image database I am dealing with is rather large, this would save me a huge amount of disk-space.

rragundez:
It is possible by setting the parameters `drop_duplicates=False`, but I recommend you to not do it the way you are doing it but to use the `class_weight` parameter of `fit_generator` to balance your classes.
Here is a blogpost example where I use the `class_weight` parameter  https://blog.godatadriven.com/rod-keras-multi-label

Issue #96, opened 2018-11-21
ImageDataGenerator.flow_from_dataframe keeps loading when directory has subdirectories


fahadakhan96:
I'm working on the [MURA dataset](https://stanfordmlgroup.github.io/competitions/mura/) by Stanford. I'm trying to load the dataset using Keras's ImageDataGenerator. The data is in the following hierarchy:

[![The directory hierarchy][1]][1]

The `study1_positive` folder contains the images.

`ImageDataGenerator.flow_from_directory` cannot be used with this folder structure, therefore I tried using the `flow_from_dataframe` method. 

However, when run, the code keeps on executing and doesn't stop.

Following is the format of the Pandas DataFrame that I'm passing to the `flow_from_directory` method:

[![The DataFrame passed to flow_from_dataframe][2]][2]

I've also tried changing the labels to 'abnormal' and 'normal' in place of 1 and 0, respectively.

Below is the code:

    train_imggen = ImageDataGenerator(rescale=1./255, rotation_range=30,
                                  horizontal_flip=True)

    train_loader = train_imggen.flow_from_dataframe(traindf, './', shuffle=True,
                                                x_col='path', y_col='label',
                                                color_mode='grayscale',
                                                target_size=(320,320), 
                                                class_mode='binary', 
                                                batch_size=8)

  [1]: https://i.stack.imgur.com/ZZVs6.png
  [2]: https://i.stack.imgur.com/QNeie.png

fahadakhan96:
Thanks, @smurak!

Your fix worked! Didn't need Step 4, though.

Please feel free to close this issue.

arminf82:
Thanks, @smurak! Great

It is working now with absolute paths! Step 4. kind is not working on my system.

yasar-rehman:
Here is my code using absolute path: @Vijayabhaskar96  

 train_df = pd.DataFrame(train_img_data)
    train_df.columns = ['id', 'label']

    test_df = pd.DataFrame(test_img_data)
    test_df.columns = ['id', 'label']

    print(train_df['id'][0])
    print('******************************************************')
  

    datagen = ImageDataGenerator(rescale=1./255)

    train_generator = datagen.flow_from_dataframe(train_df, None,
        x_col='id',
        y_col='label',
        has_ext=True,
        batch_size=args.batch_size,
        seed=42,
        shuffle=True,
        class_mode="sparse",
        target_size=(224,224),
        color_mode='rgb',
        interpolation='nearest'
    )


[...]

PR #182, opened 2019-03-08
Remove the parameter drop_duplicates

rragundez:
### Summary
Currently the `drop_diuplicates` parameter from `flow_from_dataframe`and ``DataFrameIterator` can deduplicate filenames from the user's input. My arguments for removing it are:

1.- In the same way that `sort` and `has_ext` have nothing to do with the logic the user expects and that should be handled by the `DataFrameIterator`, `drop_duplicates` also has nothing to do with the logic the class should handle. Look #123 and #122 for more explanation.
2.- It produces an unexpected behavior to the user, and several issues have been raised already because of this: #96 #175 and #181

### Related Issues
 #96 #175  #181

### PR Overview

### DIFF
--- a/keras_preprocessing/image/image_data_generator.py
+++ b/keras_preprocessing/image/image_data_generator.py
@@ -556,7 +556,6 @@ class ImageDataGenerator(object):
                             save_format='png',
                             subset=None,
                             interpolation='nearest',
-                            drop_duplicates=True,
                             validate_filenames=True,
                             **kwargs):
         """Takes the dataframe and the path to a directory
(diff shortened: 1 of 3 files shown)


keras_preprocessing/image/image_data_generator.py
```diff
@@ -652,6 +649,9 @@ def flow_from_dataframe(self,
             warnings.warn('sort is deprecated, batches will be created in the'
                           'same order than the filenames provided if shuffle'
                           'is set to False.')
+        if 'drop_duplicates' in kwargs:
+            warnings.warn('drop_duplicates is deprecated.', DeprecationWarning)
```
Dref360: The user would probably like some indications: How can they remove duplicate themselves? Using df.drop_duplicates.

LGTM otherwise

rragundez: done

PR #189, opened 2019-03-19
Allow numpy.array as classes in DataFrameIterator

kennethtxytqw:
### Summary
Numpy array not allowed to be used as `classes` in `DataFrameIterator`.

I understand that according to the docs:
Argument `classes`: Optional list of strings, classes to use (e.g. `["dogs", "cats"]`).
I wonder if we could expand it to include `numpy.array`?

Usually, I would derive my list of classes using `dataset['Category'].unique()` where `dataset` is a `pandas.DataFrame`, this returns a `numpy.array`.

A `numpy.array` will cause `ValueError` due to `numpy.array`'s boolean operation. Below is a screenshot of a small example:
image

P.S. I am not sure if this is intended and I am a novice in open source, so please bear with me.

### PR Overview

### DIFF
--- a/keras_preprocessing/image/dataframe_iterator.py
+++ b/keras_preprocessing/image/dataframe_iterator.py
@@ -44,7 +44,7 @@ class DataFrameIterator(BatchFromFilesMixin, Iterator):
         target_size: tuple of integers, dimensions to resize input images to.
         color_mode: One of `"rgb"`, `"rgba"`, `"grayscale"`.
             Color mode to read images.
-        classes: Optional list of strings, classes to use (e.g. `["dogs", "cats"]`).
+        classes: Optional numpy array/list of strings, classes to use (e.g. `["dogs", "cats"]`).
             If None, all classes in `y_col` will be used.
         class_mode: one of "binary", "categorical", "input", "multi_output",
             "raw", "sparse" or None. Default: "categorical".
@@ -185,7 +185,7 @@ class DataFrameIterator(BatchFromFilesMixin, Iterator):
                                 .format(self.class_mode, y_col))
         # check that if binary there are only 2 different classes
         if self.class_mode == 'binary':
-            if classes:
+            if classes is not None:
                 classes = set(classes)
                 if len(classes) != 2:
                     raise ValueError('If class_mode="binary" there must be 2 '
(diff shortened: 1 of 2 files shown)


keras_preprocessing/image/dataframe_iterator.py
```diff
@@ -202,7 +202,7 @@ def _check_params(self, df, x_col, y_col, weight_col, classes):
                                 'values must be type string, list or tuple.'
                                 .format(self.class_mode, y_col))
         # raise warning if classes are given but will be unused
-        if classes and self.class_mode in {"input", "multi_output", "raw", None}:
+        if classes != None and self.class_mode in {"input", "multi_output", "raw", None}:
```
Dref360: Please use classes is not None

keras_preprocessing/image/dataframe_iterator.py
```diff
@@ -44,7 +44,7 @@ class DataFrameIterator(BatchFromFilesMixin, Iterator):
         target_size: tuple of integers, dimensions to resize input images to.
         color_mode: One of `"rgb"`, `"rgba"`, `"grayscale"`.
             Color mode to read images.
-        classes: Optional list of strings, classes to use (e.g. `["dogs", "cats"]`).
+        classes: Optional numpy array/list of strings, classes to use (e.g. `["dogs", "cats"]`).
```
Dref360: All Iterable would work right?

kennethtxytqw: Yes! Should I change it to "Optional `Iterable` with strings, ..."?

keras_preprocessing/image/dataframe_iterator.py
kennethtxytqw: Done! Thank you for your feedback!

kennethtxytqw:
@Dref360 Added unit test.

keras_preprocessing/image/dataframe_iterator.py
```diff
@@ -185,7 +185,7 @@ def _check_params(self, df, x_col, y_col, weight_col, classes):
                                 .format(self.class_mode, y_col))
         # check that if binary there are only 2 different classes
         if self.class_mode == 'binary':
-            if classes:
+            if classes is not None:
```
kennethtxytqw: Thanks to the unit test. I realized other parts require it as well.

keras_preprocessing/image/dataframe_iterator.py
```diff
@@ -235,7 +235,7 @@ def remove_classes(labels, classes):
                     .format(type(labels), y_col)
                 )
 
-        if classes:
+        if classes is not None:
```
kennethtxytqw: Here too!

kennethtxytqw:
Travis CI build failed due to configuration, is there anything I can do about it?

rragundez:
@kennethtxytqw you could derive your list of classes by `dataset['Category'].unique().tolist()` instead of `dataset['Category'].unique()`, just by adding `tolist()`. No need for a numpy array in my opinion.

kennethtxytqw:
Thanks @rragundez for the solution, I actually did that in the end. 

It is just that I expected the numpy array to be treated the same way and when i got the error it to me a while to realize what is going on. 

I will go ahead and close this then.

PR #193, opened 2019-03-28
Feature/fix performance issue

Norbi1986:
### Summary
We noticed a performance issue with the latest release(1.0.9). 
We did a root cause analysis and we found the source of issue.
The "[os.path.join(self.directory, fname) for fname in self.filenames]" was refactorized into @property function and had been called with each iteartion by _get_batches_of_transformed_samples() function:
e.g: 100(iteration) x 155.096(number of iamges) = 15.509.600(path generated)

I moved the filepath generator to the _init_() function from the @property function
So, the file_paths will be generated just one time with _init()_ function

### PR Overview

### DIFF
--- a/keras_preprocessing/image/dataframe_iterator.py
+++ b/keras_preprocessing/image/dataframe_iterator.py
@@ -121,7 +121,7 @@ class DataFrameIterator(BatchFromFilesMixin, Iterator):
                                                             subset,
                                                             interpolation)
         df = dataframe.copy()
-        self.directory = directory
+        self.directory = directory or ''
         self.class_mode = class_mode
         self.dtype = dtype
         # check that inputs match the required class_mode
@@ -157,14 +157,13 @@ class DataFrameIterator(BatchFromFilesMixin, Iterator):
         else:
             print('Found {} {} image filenames belonging to {} classes.'
                   .format(self.samples, validated_string, num_classes))
+        self._filepaths = [
+            os.path.join(self.directory, fname) for fname in self.filenames
+        ]
         super(DataFrameIterator, self).__init__(self.samples,
                                                 batch_size,
                                                 shuffle,
                                                 seed)
-        root = self.directory or ''
-        self.path_to_files = [
-            os.path.join(root, fname) for fname in self.filenames
-        ]
 
     def _check_params(self, df, x_col, y_col, weight_col, classes):
         # check class mode is one of the currently supported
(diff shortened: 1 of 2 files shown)


datumbox:
This patch improves significantly the speed. From 1.0.5 to 1.0.9 the execution times increased by 5% due to the refactoring. This patch reduced the execution times comparing to 1.0.9 by 12%.

keras_preprocessing/image/dataframe_iterator.py
```diff
@@ -161,6 +161,10 @@ def __init__(self,
                                                 batch_size,
                                                 shuffle,
                                                 seed)
+        root = self.directory or ''
+        self.path_to_files = [
```
rragundez: Can you call the attribute `self.filepaths`? It is a common practice to add a prefix underscore if the attribute will be exposed via a property, thanks. 
Perhaps we do not need to create that intermediate `root` variable.
and can you change this logic above the instantiation of the parent class? Line 159

keras_preprocessing/image/directory_iterator.py
```diff
@@ -140,9 +140,13 @@ def __init__(self,
...
                                                 seed)
 
+        self.path_to_files = [
+            os.path.join(self.directory, fname) for fname in self.filenames
+        ]
+
     @property
     def filepaths(self):
-        return [os.path.join(self.directory, fname) for fname in self.filenames]
+        return self.path_to_files
```
rragundez: Can you call the attribute `self.filepaths`? It is a common practice to add a prefix underscore if the attribute will be exposed via a property, thanks. 
Perhaps we do not need to create that intermediate `root` variable.
and can you change this logic above the instantiation of the parent class? Line 138

keras_preprocessing/image/dataframe_iterator.py
Norbi1986: Yes, I fixed my code according to your remark

keras_preprocessing/image/directory_iterator.py
Norbi1986: Yes, I fixed my code according to your remark

keras_preprocessing/image/directory_iterator.py
```diff
@@ -90,7 +90,7 @@ def __init__(self,
                                                             save_format,
                                                             subset,
                                                             interpolation)
-        self.directory = directory
+        self.directory = directory or ''
```
rragundez: In `DirectoryIterator` a the `directory` argument is not optional, no need for the `or` operator.

Norbi1986: Yes, you are right.
I fixed it

rragundez:
Thanks for your contribution! LGTM

PR #198, opened 2019-04-17
Delete test_documentation.py

rragundez:
### Summary
Remove unused test for documentation. Keras preprocessing does not manage documentation itself.


Dref360:
Even if we don't manage the doc, I think it is valuable to keep the same document format as Keras.

rragundez:
I think it might be valuable to also test documentation, similar to: https://github.com/keras-team/keras/blob/master/tests/docs/test_documentation.py
What do you think?

Dref360:
I agree, do you want to do it? Otherwise I'll do it this week.

rragundez:
Would be great if you can give me a hand. If there are many test failing because of this I can help shaping the docs to what it supposed to be :), I'll close this PR then.

Issue #203, opened 2019-04-22
May raise modulo by zero error


siddheshsathe:
Please make sure that the boxes below are checked before you submit your issue.
If your issue is an **implementation question**, please ask your question on [StackOverflow](http://stackoverflow.com/questions/tagged/keras) or [on the Keras Slack channel](https://keras-slack-autojoin.herokuapp.com/) instead of opening a GitHub issue.

Thank you!

- [x] Check that you are up-to-date with the master branch of keras-preprocessing. You can update with:
`pip install git+git://github.com/keras-team/keras-preprocessing.git --upgrade --no-deps`

- [x] Provide a link to a GitHub Gist of a Python script that can reproduce your issue (or just copy the script here if it is short).

**Source:** https://github.com/keras-team/keras-preprocessing/blob/6ff9af6971f97f69c39d090be15c2f7d93aa4d6d/keras_preprocessing/image/iterator.py#L85

- Under class `Iterator`, if the initialization happens with 0 number of images for any number of class, it may raise `modulo by zero error`
- Solution would be to assert if `n == 0` while initializing the `Iterator` class.

rragundez:
Solved at #204

PR #204, opened 2019-04-22
Fix ZeroDivisionError issue while iterating over directory containing 0 images

siddheshsathe:
### Summary
While iterating over directory containing 0 images, it raises ZeroDivisionError. Asserting before modulo by zero.
### Related Issues
#203
### PR Overview

### DIFF
--- a/keras_preprocessing/image/iterator.py
+++ b/keras_preprocessing/image/iterator.py
@@ -34,9 +34,6 @@ class Iterator(IteratorType):
     white_list_formats = ('png', 'jpg', 'jpeg', 'bmp', 'ppm', 'tif', 'tiff')
 
     def __init__(self, n, batch_size, shuffle, seed):
-        if n == 0:
-                raise FileNotFoundError('No files found under provided \
-                    directory to iterate over')
         self.n = n
         self.batch_size = batch_size
         self.seed = seed
@@ -85,6 +82,9 @@ class Iterator(IteratorType):
             if self.batch_index == 0:
                 self._set_index_array()
 
+            if self.n == 0:
+                raise FileNotFoundError('No files found under provided \
+                directory to iterate over')
             current_index = (self.batch_index * self.batch_size) % self.n
             if self.n > current_index + self.batch_size:
                 self.batch_index += 1
--- a/keras_preprocessing/image/iterator.py
+++ b/keras_preprocessing/image/iterator.py
@@ -31,7 +31,7 @@ class Iterator(IteratorType):
         shuffle: Boolean, whether to shuffle the data between epochs.
         seed: Random seeding for data shuffling.
     """
-    white_list_formats = {'png', 'jpg', 'jpeg', 'bmp', 'ppm', 'tif', 'tiff'}
+    white_list_formats = ('png', 'jpg', 'jpeg', 'bmp', 'ppm', 'tif', 'tiff')
 
     def __init__(self, n, batch_size, shuffle, seed):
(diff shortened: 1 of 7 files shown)


keras_preprocessing/image/iterator.py
```diff
@@ -82,6 +82,7 @@ def _flow_index(self):
             if self.batch_index == 0:
                 self._set_index_array()
 
+            assert self.n != 0, "Won't iterate over 0 number of images"
```
rragundez: The best would be to raise an Exception.

rragundez:
Please add unit tests

keras_preprocessing/image/iterator.py
siddheshsathe: - Moved this to __init__
- Added `FileNotFoundError`

siddheshsathe: - Moved the same to `_flow_index` method of `Iterator` class
- Raising `Exception` since CI tool gave `Unknown: FileNotFoundError`

rragundez:
Thinking a bit more about it I do not think it should raise an exception. It should just do what is suppose to do, iterate, which means returning zero samples on each iteration. A message that there are 0 samples/observations found is already given by the methods in `ImageDataGenerator`.  See inline comments for proposal fix.

keras_preprocessing/image/iterator.py
rragundez: instead of raising on exception it should just iterate returning zero sample on each iteration. This can be accomplish by something like:
```
            if self.n == 0:
                current_index = 0
            else:
                current_index = (self.batch_index * self.batch_size) % self.n 
```

siddheshsathe: Agreed over the suggestion.

keras_preprocessing/image/iterator.py
```diff
@@ -82,7 +82,11 @@ def _flow_index(self):
             if self.batch_index == 0:
                 self._set_index_array()
 
-            current_index = (self.batch_index * self.batch_size) % self.n
+            if self.n == 0:
+                # Avoiding Division by zero error
```
rragundez: `Avoiding modulo by zero`

tests/image_test.py
```diff
@@ -1358,6 +1358,10 @@ def test_validate_filename(self, tmpdir):
         assert not image.validate_filename(str(filename), valid_extensions)
         assert not image.validate_filename('some_test_file.png', valid_extensions)
 
+    def test_iterator_empty_directory(self):
+        iterator = image.Iterator(0, 0, False, 0)
+        next(iterator.index_generator)
```
rragundez: test that there a no samples returned by the iterator., add also test with a different batch size

tests/image_test.py
```diff
@@ -1359,8 +1359,11 @@ def test_validate_filename(self, tmpdir):
         assert not image.validate_filename('some_test_file.png', valid_extensions)
 
     def test_iterator_empty_directory(self):
-        iterator = image.Iterator(0, 0, False, 0)
-        next(iterator.index_generator)
+        # Testing with randomly chosen batch sizes
```
rragundez: The batches are not generated randomly, something like: "Testing with different batch sizes" is a best description. Also there is no need to check with more batch sizes, only 0 and non-zero.

siddheshsathe: Done.

rragundez:
Thanks for you collaboration!

PR #206, opened 2019-04-28
Refactor tests of the image module

rragundez:
### Summary
Refactor tests of the 'image' submodule to match the skeleton of the submodule itself. Now instead of having a single file with all the tests, some logic has been separated into different files according to the class or function being tested. This allows the addition and debugging of failed tests ti be much simpler and faster.

### PR Overview


tests/image/dataframe_iterator_test.py
```diff
@@ -0,0 +1,653 @@
...
+    batch = next(iterator)
+    assert len(batch) == 2
+    assert isinstance(batch[0], np.ndarray)
+    assert isinstance(batch[1], np.ndarray)
+    generator = image_data_generator.ImageDataGenerator()
+    df_iterator = generator.flow_from_dataframe(df, x_col='filepaths')
+    df_iterator_dir = generator.flow_from_dataframe(df, str(tmpdir))
+    df_sparse_iterator = generator.flow_from_dataframe(df, str(tmpdir),
+                                                       class_mode="sparse")
+    if np.isnan(df_sparse_iterator.classes).any():
```
Dref360: When does this can happen?

tests/image/dataframe_iterator_test.py
```diff
@@ -0,0 +1,653 @@
...
+    generator = image_data_generator.ImageDataGenerator()
+    df_autoencoder_iterator = generator.flow_from_dataframe(df, str(tmpdir),
+                                                            x_col="filename",
+                                                            y_col=None,
+                                                            class_mode="input")
+
+    batch = next(df_autoencoder_iterator)
+
+    # check if input and output have the same shape and they're the same
+    assert(batch[0].all() == batch[1].all())
```
Dref360: The `.all` is misplaced I think.

`np.allclose(batch[0], batch[1])`

tests/image/dataframe_iterator_test.py
```diff
@@ -0,0 +1,653 @@
...
+        assert np.array_equal(a1, a2)
+        assert np.array_equal(a1, a3)
+        assert np.array_equal(c1, c2)
+        assert np.array_equal(c1, c3)
+
+
+def test_dataframe_iterator_n(all_test_images, tmpdir):
+
+    # save the images in the tmpdir
+    count = 0
```
Dref360: I'm not sure of the value of this test?

rragundez:
I saw you comments inline @Dref360 , basically I just splitted the current tests that were already there.  I would try to address your questions though.

tests/image/dataframe_iterator_test.py
rragundez: I don't know. I just moved a test that was already there. But I converted into an assertion

tests/image/dataframe_iterator_test.py
rragundez: changed

tests/image/dataframe_iterator_test.py
rragundez: I'm also not sure, it was already there. @Vijayabhaskar96  do you know? I think you originally made  this one.

Vijayabhaskar96: Looks like It matches the number of filenames found and the number of files in the dataframe is the same. @rragundez I think I didn't author this.

rragundez: thanks @Vijayabhaskar96

rragundez:
I've change the PR taking into account the comments @Dref360

PR #207, opened 2019-05-02
Tests for documentation

Dref360:
### Summary
Added tests for documentation.
### PR Overview

### DIFF
--- a/keras_preprocessing/image/dataframe_iterator.py
+++ b/keras_preprocessing/image/dataframe_iterator.py
@@ -108,7 +108,6 @@ class DataFrameIterator(BatchFromFilesMixin, Iterator):
                  save_format='png',
                  subset=None,
                  interpolation='nearest',
-                 dtype='float32',
                  validate_filenames=True):
 
         super(DataFrameIterator, self).set_processing_attrs(image_data_generator,
@@ -123,7 +122,6 @@ class DataFrameIterator(BatchFromFilesMixin, Iterator):
         df = dataframe.copy()
         self.directory = directory or ''
         self.class_mode = class_mode
-        self.dtype = dtype
         # check that inputs match the required class_mode
         self._check_params(df, x_col, y_col, weight_col, classes)
         if validate_filenames:  # check which image files are valid and keep them
@@ -133,6 +131,7 @@ class DataFrameIterator(BatchFromFilesMixin, Iterator):
             num_classes = len(classes)
             # build an index of all the unique classes
             self.class_indices = dict(zip(classes, range(len(classes))))
+        # retrieve only training or validation set
         # retrieve only training or validation set
         if self.split:
             num_files = len(df)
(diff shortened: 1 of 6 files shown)


keras_preprocessing/image/dataframe_iterator.py
```diff
@@ -134,6 +135,7 @@ def __init__(self,
             # build an index of all the unique classes
             self.class_indices = dict(zip(classes, range(len(classes))))
         # retrieve only training or validation set
+        # retrieve only training or validation set
```
rragundez: the comment is duplicated

keras_preprocessing/image/utils.py
```diff
@@ -156,7 +158,7 @@ def _iter_valid_files(directory, white_list_formats, follow_links):
             containing files to be counted
         white_list_formats: Set of strings containing allowed extensions for
             the files to be counted.
-        follow_links: Boolean.
+        follow_links: Boolean, follow symbolic links to subdirectories
```
rragundez: finish the sentence with a dot

keras_preprocessing/image/utils.py
```diff
@@ -189,7 +191,7 @@ def _list_valid_filenames_in_directory(directory, white_list_formats, split,
             E.g.: `segment=(0.6, 1.0)` would only account for last 40 percent
             of images in each directory.
         class_indices: dictionary mapping a class name to its index.
-        follow_links: boolean.
+        follow_links: boolean, follow symbolic links to subdirectories
```
rragundez: finish the sentence with a dot

tests/test_documentation.py
```diff
@@ -0,0 +1,174 @@
...
+from itertools import compress
+
+import pytest
+
+modules = ['keras_preprocessing',
+           'keras_preprocessing.image',
+           'keras_preprocessing.sequence',
+           'keras_preprocessing.text']
+
+# Tokenizer is being refactored PR #106
```
rragundez: how can we be sure to remember to remove this after merging the PR?

tests/test_documentation.py
```diff
@@ -0,0 +1,174 @@
...
+    handle_function(name, member)
+
+
+def handle_module(mod):
+    for name, mem in inspect.getmembers(mod):
+        if inspect.isclass(mem):
+            handle_class(name, mem)
+        elif inspect.isfunction(mem):
+            handle_function(name, mem)
+        elif 'keras_preprocessing' in name and inspect.ismodule(mem):
```
rragundez: In this case I'm not sure we need the check `'keras_preprocessing' in name`, since I think we want to check every function and module, not only the top level exposed classes or functions. At the moment this is only relevant for the `image` module as it contains a deeper structure, but I think we should check every function and class regardless of where it is in the package tree. 
This will mean that functions or classes exposed via the `__init__.py` file will be checked double.
If we remove that condition then the comment in the line below should be removed.

Dref360: I think this is parsing the entire tree. This check is there so that we do not validate dependencies like Numpy.

rragundez:
```
import inspect
import numpy

print( [x for x in inspect.getmembers(numpy) if inspect.ismodule(x[1])]  )
```
The names in the tuples do not contain `numpy` in this case, in the same way that they won't contain `keras_preprocessing`. But you are also right that if you remove it it will also traverse all modules being imported :s. 
Perhaps indeed the best is to leave it as is. Just change the comment below for `# Only test keras_preprocessing' modules`

PR #208, opened 2019-05-05
Adding secure_image.py for supporting training on encrypted images.

akshaybahadur21:
### Summary
I have added support for encrypting directory of image and then write a generator for training the model on the encrypted model.
Adding the link to collab repo as well as a standalone package for the same.
### PR Overview

- [ yes ] This PR requires to update the documentation [y/n] (make sure the docs are up-to-date)

### DIFF
--- a/keras_preprocessing/image/__init__.py
+++ b/keras_preprocessing/image/__init__.py
@@ -8,4 +8,5 @@ from .directory_iterator import DirectoryIterator
 from .image_data_generator import ImageDataGenerator
 from .iterator import Iterator
 from .numpy_array_iterator import NumpyArrayIterator
+from .secure_image import *
 from .utils import *
--- a/keras_preprocessing/image/secure_image.py
+++ b/keras_preprocessing/image/secure_image.py
@@ -1,0 +1,143 @@
+import hashlib
+import math
+import os
+
+import numpy as np
+from PIL import Image
+from skimage.io import imread
+
+
+def transform(src_dir, dest_dir, count, block_size, image_x, image_y):
+    """
+    This function checks the directory and encrypts the image and saves it in the destination folder
+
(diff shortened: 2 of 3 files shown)


akshaybahadur21:
@rragundez - Can you please review?

rragundez:
@akshaybahadur21 thanks for your PR, i quickly reviewed it, but my question is what is this solving? Your code seems a generic function to encrypt files, and I don't see how is related to Keras functionality. The best would have been to create an issue and see if your idea gets some traction, before doing all the effort.  

For example, I don't see why Keras would care about encrypting files, reading from encrypted files maybe, which means I would think a better approach would have been to implement something in the `load_img` function. Still I would like to see that this solves a current problem in the community, which I currently don't.

akshaybahadur21:
@rragundez - Thanks for taking a look at my PR.
This functionality is for training on the cloud or data centers. You can read more about it.
I have created an add on library as well.
My idea was to create a `fit_generator` to decrypt and yield these images that the actual images are not stored on the cloud.
You can read more here : https://github.com/akshaybahadur21/keras-secure-image

rragundez:
I understand what it achieves, but I don't see why Keras would care about encrypting images. That has nothing to do with what Keras is for, and also the user should be the one responsible to encrypting the images as he sees fit. Which means no decryption function can be written since it will have to address all possible ways. Therefore the only possible addition that I might see useful is modifying `load_img`, specifically this line https://github.com/keras-team/keras-preprocessing/blob/master/keras_preprocessing/image/utils.py#L110
To allow the user to send whatever function wanted for reading images, included decryption.

akshaybahadur21:
@rragundez - Oh.
Yes, I get your point now.
Let me make the changes and raise another request.

PR #244, opened 2019-09-26
Remove scaling of images when saving to disk

### LINKED ISSUE #221: ImageDataGenerator makes colors change which I don't expect
Hey guys, I'll just state out the core problem. If I use ImageDataGenerator to read the images and then output without any change, the output image still looks different to the original ones. 
Here's my code:

img=load_img("anyimage.jpg")
img=img_to_array(img)
img=np.expand_dims(img,axis=0)
datagen = ImageDataGenerator()
k=0
for batch in datagen.flow(img, batch_size=1, save_to_dir='', save_prefix='', save_format='jpg'):
    k=k+1
    if k==1:
        break

Since all the parameters in ImageDataGenerator are set to false by default, the output image should be exactly the same as the input image. But actually the output image looks "brighter", and that can be confirmed by examining the pixel values. Could anyone help me with that? I'm greatly thankful here. 

You may wonder why I use ImageDataGenerator to do "nothing". Actually I simplify the problem. While the whole issue is, I use the ImageDataGenerator to generate "rotation, fliping and horizontal & vertical shifting" images, but the colour of all the images generated are also changed which I don't expect. So after training with these images, all the generated images pass the neural network, but the original images fail a lot. The colour change really confuses me. 

Please make sure that the boxes below are checked before you submit your issue.
If your issue is an **implementation question**, please ask your question on [StackOverflow](http://stackoverflow.com/questions/tagged/keras) or [on the Keras Slack channel](https://keras-slack-autojoin.herokuapp.com/) instead of opening a GitHub issue.

Thank you!

- [x] Check that you are up-to-date with the master branch of keras-preprocessing. You can update with:
`pip install git+git://github.com/keras-team/keras-preprocessing.git --upgrade --no-deps`

- [X] Provide a link to a GitHub Gist of a Python script that can reproduce your issue (or just copy the script here if it is short).

rragundez:
### Summary
When using `ImageDataGenerator` methods and saving to disk, the images saved to disk are scaled and do not present an accurate representation of what the generator is actually using as samples in the produced batches.

Current behavior:
```
import numpy as np
from datetime import datetime
from glob import glob

from keras_preprocessing.image.utils import load_img, array_to_img, img_to_array

x = np.random.randint(20, 30, size=(1, 100, 100 , 1))
img = array_to_img(x[0], scale=False)
prefix = str(datetime.now())
batch_gen = ImageDataGenerator().flow(x,
                                      batch_size=1,
                                      save_to_dir='.',
                                      save_prefix=prefix)
x = next(batch_gen)

img2 = load_img(glob('{}*'.format(prefix))[0], color_mode='grayscale')
x2 = img_to_array(img2)
np.all(x[0] == x2)
>>> False
```
Visually comparing the images also yield differences because of the scaling.

Selection_031

This is confusing for the user if the user wants to have a visual check. This will be specially important now that we can have batches with grayscale images of 16-bit and 32-bit pixel values, when rescaling they become 8-bit and look extremely different.

NOTE: I have the feeling that the scaling logic itself is wrong but I will address that in another issue/PR

### Related Issues
Closes #221 

### PR Overview

### DIFF
--- a/keras_preprocessing/image/iterator.py
+++ b/keras_preprocessing/image/iterator.py
@@ -241,7 +241,7 @@ class BatchFromFilesMixin():
         # optionally save augmented images to disk for debugging purposes
         if self.save_to_dir:
             for i, j in enumerate(index_array):
-                img = array_to_img(batch_x[i], self.data_format, scale=True)
+                img = array_to_img(batch_x[i], self.data_format, scale=False)
                 fname = '{prefix}_{index}_{hash}.{format}'.format(
                     prefix=self.save_prefix,
                     index=j,
--- a/keras_preprocessing/image/numpy_array_iterator.py
+++ b/keras_preprocessing/image/numpy_array_iterator.py
@@ -156,7 +156,7 @@ class NumpyArrayIterator(Iterator):
 
         if self.save_to_dir:
             for i, j in enumerate(index_array):
-                img = array_to_img(batch_x[i], self.data_format, scale=True)
+                img = array_to_img(batch_x[i], self.data_format, scale=False)
                 fname = '{prefix}_{index}_{hash}.{format}'.format(
                     prefix=self.save_prefix,
                     index=j,


rragundez:
I made a quick [blog post draft](https://medium.com/@rragundez/medical-images-now-supported-by-keras-imagedatagenerator-e67d1c2a1103)  where this difference is clearly showed when saving images to disk. I will publish once this one is merged.

Dref360:
I think we just need to fix the rescaling. Not sure what was the initial intention, but
`x + max(-np.min(x), 0)` should be `x - np.min(x)`

rragundez:
I already made an issue for this and have the branch in my fork for the PR. But this something different. This is saving to disk an images that suppose to represent the ones being used by the model but they are not as they are scaled by default. With others but specially with 16-bit and 32-bit images this is a problem.

Dref360:
We have no guarantee that the image is made of ints right?

rragundez:
It doesn't have to, if it is float32 it can still contain the information that a 16-bit and 32-bit int image will have. But then if when saving to disk is scaled then it will lose the information while is actually being used correctly internally with the latest changes. That's what #221 is experiencing.

Dref360:
Okay I get what is the problem, but I don't think this is the right solution.

If we normalize the input between [-1, 1]. Without the scale argument, we will cast it to `uint8` before making it an image and we will lose information and get a black image.

As stated in the doc, `save_to_dir` is useful for visualizing what you are doing.

rragundez:
got it! indeed this is not the correct solution for that case.

rragundez:
I currently do not see any other solution other than make an exception for scaling images with pixels values above 255. Or even better but more work, add color modes `grayscale_16bit` and `grayscale_32bit`, then specifically we know how to treat these images.

bittergourd1224:
Hi there. According to your discussion, is how the saved image is processed related to the format of the input image? For me, an option that can turn the scaling on or off when saving images after generating is enough. Then with the option off, images saved will remain the same as the ones after augmentation.

PR #246, opened 2019-09-27
Update rescaling logic

### LINKED ISSUE #245: Scaling logic is weird / arbitrary
The scaling logic in [`array_to_img`](https://github.com/keras-team/keras-preprocessing/blob/master/keras_preprocessing/image/utils.py#L265) seems to be a bit strange.

I was expecting this parameter to always rescale the pixel values such that the minimum vaalue is 0 and the  maximum value is 255. That is consistent with best practices scaling of pixel values. Instead the logic looks like:
```
        x = x + max(-np.min(x), 0)
        x_max = np.max(x)
        if x_max != 0:
            x /= x_max
        x *= 255
```
This produces weird and in my opinion unexpected results:

- if any pixel value is negative it will indeed scale the image with minimum value of 0 and maximum of 255.
- if no pixel value is negative then it will scale the image such that the minimum value remains the same but the maximum is 255... this losses any proportionality or intuition on what is doing.

I was going to make a PR directly butt perhaps there is some reason I am not seeing, even though the PR is a small  change. Specifically:
`x = x + max(-np.min(x), 0)` for `x = x - np.min(x)`

rragundez:
### Summary
This PR corrects the scaling logic.

### Related Issues
Closes #245 

### PR Overview

### DIFF
--- a/keras_preprocessing/image/utils.py
+++ b/keras_preprocessing/image/utils.py
@@ -234,7 +234,8 @@ def array_to_img(x, data_format='channels_last', scale=True, dtype='float32'):
         x: Input Numpy array.
         data_format: Image data format, either "channels_first" or "channels_last".
             Default: "channels_last".
-        scale: Whether to rescale image values to be within `[0, 255]`.
+        scale: Whether to rescale the image such that minimum and maximum values
+            are 0 and 255 respectively.
             Default: True.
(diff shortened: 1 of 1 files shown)


rragundez:
@Dref360 this is the change to the scaling logic we discussed.

PR #251, opened 2019-10-11
Add inter_batch_stride param to TimeSeriesGenerator

rjmccabe3701:
### Summary

Grant explicit control for inter-batch sample ordering in TimeSeriesGenerator.
This change is backwards compatible.

### Related Issues

See this [stackoverflow](https://stackoverflow.com/questions/58276337/proper-way-to-feed-time-series-data-to-stateful-lstm) question for background.

### PR Overview

The original TimeSeriesGenerator emits batches where the first sample is
a stride ahead of the previous batch's last sample. Many times more
control is required (especially if feeding to a stateful RNN).

The inter_batch_stride option allows user to explicitly specify the
inter-batch stride relationships.  If this option is None (or not
supplied), the original behavior is maintained.

- [y] This PR changes the current API [y/n] (all API changes need to be approved by fchollet)

### DIFF
--- a/keras_preprocessing/sequence.py
+++ b/keras_preprocessing/sequence.py
@@ -278,8 +278,16 @@ class TimeseriesGenerator(object):
             `data[i]`, `data[i-r]`, ... `data[i - length]`
             are used for create a sample sequence.
         stride: Period between successive output sequences.
-            For stride `s`, consecutive output samples would
-            be centered around `data[i]`, `data[i+s]`, `data[i+2*s]`, etc.
+            For stride `Stride`, consecutive output samples would
+            be centered around `data[i]`, `data[i+Stride]`, `data[i+2*Stride]`, etc.
+            If ``inter_batch_stride`` is None, then `Stride` carries over
+            between batches: the first sample of a batch is a stride ahead of the
+            last sample of the previous batch. The formula for the collection of
+            samples in a batch when ``inter_batch_stride = None`` is
+            ```[[[data[batch_size*Stride*b + s*Stride + l]
+                    for l in range(length)] for s in range(batch_size)]
+                    for b in range(num_batches)]
+            ```
         start_index: Data points earlier than `start_index` will not be used
             in the output sequences. This is useful to reserve part of the
             data for test or validation.
@@ -292,6 +300,10 @@ class TimeseriesGenerator(object):
             in reverse chronological order.
         batch_size: Number of timeseries samples in each batch
             (except maybe the last one).
+        inter_batch_stride: If not None, grants explicit control w.r.t. the
+            inter-batch stride relationship -- instead of the first sample
+            of a batch being a stride ahead of the last sample in the previous
+            batch.
 
     # Returns
         A [Sequence](/utils/#sequence) instance.
@@ -305,6 +317,8 @@ class TimeseriesGenerator(object):
     data = np.array([[i] for i in range(50)])
     targets = np.array([[i] for i in range(50)])
 
+    ####
+    #Test 1
     data_gen = TimeseriesGenerator(data, targets,
                                    length=10, sampling_rate=2,
                                    batch_size=2)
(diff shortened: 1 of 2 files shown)


OverLordGoldDragon:
I suggest staying clear of `TimeseriesGenerator` altogether; I'm not too familiar with it as I never used it, but I did so for a reason; consider:

```python
import numpy as np

data = np.random.randn(240, 4)  # (timesteps, channels)
length = 6
batch_size = 8
stride = 1
start_index, end_index, index = 6, 40, 0
```
```python
i = start_index + batch_size * stride * index
rows = np.arange(i, min(i + batch_size *
						stride, end_index + 1), stride)

samples = np.array([data[row - length:row:sampling_rate]
					for row in rows])
print(samples.shape)
# (8, 6, 4)
```
Batch dimension and a feature dimension (timesteps) are _mixed_, which is a big no-no and will obliterate training. The only workaround is if you feed data compatible with such a manipulation to begin with - good luck with that. There are other pitfalls I'll spare listing - bottom line, write your own timeseries generator.

rjmccabe3701:
> I suggest staying clear of `TimeseriesGenerator` altogether; I'm not too familiar with it as I never used it, but I did so for a reason; consider:
> 
> ```python
> import numpy as np
> 
> data = np.random.randn(240, 4)  # (timesteps, channels)
> length = 6
> batch_size = 8
> stride = 1
> start_index, end_index, index = 6, 40, 0
> ```
> 
> ```python
> i = start_index + batch_size * stride * index
> rows = np.arange(i, min(i + batch_size * stride, end_index + 1), stride)
> 
> samples = np.array([data[row - length:row:sampling_rate] for row in rows])
> print(samples.shape)
> # (8, 6, 4)
> ```
> 
> Batch dimension and a feature dimension (timesteps) are _mixed_, which is a big no-no and will obliterate training. The only workaround is if you feed data compatible with such a manipulation to begin with - good luck with that. There are other pitfalls I'll spare listing - bottom line, write your own timeseries generator.

rjmccabe3701:
@OverLordGoldDragon I understand your concerns, but this is a well-respected library.  If you feel strongly about it, help make it more understandable/extensible.

OverLordGoldDragon:
@rjmccabe3701 Keras' PR page is piled with fine improvements or extensions of functionality that are never merged - TensorFlow's a lot less phobic to this end. If I were to make a PR, pretty sure it'd be rejected or left collecting electron dust - as it'd be a total overhaul of `TimeseriesGenerator` (which it does appear to need).

For your specific instance, you need not take my word for anything - just compare the data you feed in with that which it spits out and see if it makes sense.

TanguyUrvoy:
Hello,

"Batch dimension and a feature dimension (timesteps) are mixed"
Indeed, this called *online learning* and this is precisely what time-series are about: suppose you want to train a temperature predictor from four sensors on your house, the only data you get is a single (infinite) sequence of time-indexed vectors of dimension 4. I will be quite difficult to train your model without "mixing time and batches".

You could learn purely *online* in order to keep a potentially unbounded history, I tried something like that with the (experimental) `stateful` option, but it seems more efficient (especially with GPUs) to generate fixed-length history samples of the series with a sliding/overlapping window.
If the overlapping, and the correlation it induce, is a problem for your setting you may subsample the windows. A simple way to do that is to set the `stride` parameter to a value larger than the window's length with the new version of the `TimeseriesGenerator`. But on my experience it doesn't help so much as it rejects a lot of data.

Regards,

TU

OverLordGoldDragon:
@TanguyUrvoy Misleading information; mixing batch and features dimensions is _never_ beneficial.

What you describe for online learning is _splitting the sequence_ into finite parts, then treating them as _independent_. This no longer qualifies for 'mixing' dimensions, as sequences are assumed independent to begin with. The problem arises when you have a single long sequence that is _not_ to be treated as independent, but still needs to be split up - now mixing dimensions will treat a single, causally contiguous sequence as independent sequences, and will prohibit learning any time dependence. 

The one case that differs is one-step prediction, where one sequence is fed with varying lags in a single batch - but that already accounts for above in setting up data as independent sequences. OP's case concerns statefulness, for which I little-to-no use for `TimeseriesGenerator` - [details here](https://stackoverflow.com/questions/58276337/proper-way-to-feed-time-series-data-to-stateful-lstm/58277760#58277760)

TanguyUrvoy:
@OverLordGoldDragon well I am happy that we both agree on the fact that, splitting long sequences into finite ones is a necessary evil :)

The (experimental) `stateful`option was an attempt to interleave sequences in consecutive batches in order to render `stateful` learning possible with batches of size greater than one. But I am never using it and I am not sure it really works as it is.

TU

Issue #218, opened 2019-06-10
No augmentation on validation split


Seanspt:
When using `flow_from_directory` with subset training and validation,  is it more reasonable to perform ONLY resize and rescale on validation images?

rragundez:
You can set that up with the arguments in `flow_from_directory`.

cbernet:
Hello, I'm sorry to revive this, but I am also hitting this problem. Clearly, I don't want augmentation on the validation dataset because:

* this would underestimate the accuracy.
* I do not want to have artificial fluctuations in the validation accuracy due to random augmentation.

I've studied the code for a while, and I do not see how to disable augmentation in `flow_from_directory` for the validation subset only. 

As far as I can see the augmentation is done [here](https://github.com/keras-team/keras-preprocessing/blob/master/keras_preprocessing/image/iterator.py#L238) and I can't see how this could depend on the subset : `validation` or `training`. 

Could you please tell me exactly how to disable augmentation for the validation subset in `flow_from_directory`? If you confirm that the feature is missing, and if you accept PRs, I'd be happy to contribute. 

Cheers!

rragundez:
Indeed you can create a PR: basically you need something like `... and self.subset == 'training:'` 

https://github.com/keras-team/keras-preprocessing/blob/master/keras_preprocessing/image/iterator.py#L236

Have you looked at other issues that might be related? Perhaps other people have a reason to perform augmentation in the validation set as well.

cbernet:
Thanks for your answer. Yes, something like this would work. It's actually quite simple. 

I'm not sure why anybody would like to perform augmentation on the validation set, but it's indeed possible that people might have good reasons to do so. In fact, what I would like to do is to keep the current behaviour (augmentation on the validation set is ON), and allow users to disable augmentation for this set with an additional  parameter, if you agree. And you can decide to change the default later on if needed. 

I would like to also make a pull request to the documentation to explain this, if that's ok with you. 

If you agree, I'll get started. Thanks again!

PR #256, opened 2019-10-25
can now disable augmentation

cbernet:
### Summary

Hi, this PR makes it possible to disable data augmentation. 

This feature can be needed for the validation subset for two reasons:

- Random data augmentation degrades the validation subset, which means that the true validation accuracy is in fact higher than the one currently obtained from keras. For instance, in machine learning challenges, the test dataset is not augmented. Also, in real life, we apply our models to datasets that are not augmented.  
- Due to augmentation, the validation subset changes at each epoch which leads to random fluctuations in the validation accuracy as a function of the epoch.

A new parameter has been added to the `flow` methods, `apply_augmentation`. This parameter is optional, and set to `True` by default, so that the current interface and behaviour is preserved. It can affect either the full set, the validation subset, or the training subset, depending on the subset argument. 

In issue #218  we discussed the possibility to introduce a parameter to disable augmentation for the validation set only. However, after a detailed look at the code, I concluded that it would be more logical to introduce this feature for both subsets in a generic way. 

The main reason for this decision is that a parameter like `disable_augmentation_for_validation` would have been rather weird, and more difficult to explain in the documentation to the user.  

This has the additional advantage of making it possible to disable augmentation completely for the training subset as well with a single line, e.g. when comparing augmentation vs no augmentation. 

Please let me know what you think, I'll be happy to iterate on this.

Thanks!

### Related Issues

#218 

### PR Overview

- [ n] This PR changes the current API **(I assume that new optional parameters are not considered a change to the API, but that's obviously your call)**

### DIFF
--- a/keras_preprocessing/image/dataframe_iterator.py
+++ b/keras_preprocessing/image/dataframe_iterator.py
@@ -74,6 +74,9 @@ class DataFrameIterator(BatchFromFilesMixin, Iterator):
             (if `save_to_dir` is set).
         subset: Subset of data (`"training"` or `"validation"`) if
             validation_split is set in ImageDataGenerator.
+        apply_augmentation: Boolean, can be set to False to disable
+            data augmentation, e.g. for the validation subset.
+            Standardization is still applied.
         interpolation: Interpolation method used to resample the image if the
             target size is different from that of the loaded image.
             Supported methods are "nearest", "bilinear", and "bicubic".
@@ -108,6 +111,7 @@ class DataFrameIterator(BatchFromFilesMixin, Iterator):
                  save_prefix='',
                  save_format='png',
                  subset=None,
+                 apply_augmentation=True,
                  interpolation='nearest',
                  dtype='float32',
                  validate_filenames=True):
@@ -120,6 +124,7 @@ class DataFrameIterator(BatchFromFilesMixin, Iterator):
                                                             save_prefix,
                                                             save_format,
                                                             subset,
+                                                            apply_augmentation,
                                                             interpolation)
         df = dataframe.copy()
         self.directory = directory or ''
--- a/keras_preprocessing/image/directory_iterator.py
+++ b/keras_preprocessing/image/directory_iterator.py
@@ -54,6 +54,9 @@ class DirectoryIterator(BatchFromFilesMixin, Iterator):
         follow_links: boolean,follow symbolic links to subdirectories
         subset: Subset of data (`"training"` or `"validation"`) if
             validation_split is set in ImageDataGenerator.
+        apply_augmentation: Boolean, can be set to False to disable
+            data augmentation, e.g. for the validation subset.
+            Standardization is still applied.
         interpolation: Interpolation method used to resample the image if the
             target size is different from that of the loaded image.
             Supported methods are "nearest", "bilinear", and "bicubic".
@@ -80,6 +83,7 @@ class DirectoryIterator(BatchFromFilesMixin, Iterator):
                  save_format='png',
                  follow_links=False,
                  subset=None,
+                 apply_augmentation=True,
                  interpolation='nearest',
                  dtype='float32'):
         super(DirectoryIterator, self).set_processing_attrs(image_data_generator,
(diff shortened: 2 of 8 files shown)


cbernet:
There might be something wrong with CI : the checks that do not pass seem to have nothing to do with my changes.

Dref360:
This seems related to your changes: https://travis-ci.org/keras-team/keras-preprocessing/jobs/602635040#L1097

cbernet:
@Dref360 thanks, I had missed that, and this made me realize that I need to ensure python 2.X compatibility. This should be ok now, let's see.

cbernet:
I can confirm that on my machine, the test suite passes for python 2.7 and python 3.7, and that flake8 reports 0 violations. 
This time, one of the tests is cancelled because it took more than 10 minutes. Would you know what's going on?

Dref360:
Sorry for the unacceptable delay. I got caught up at work.

To disable augmentation can't we just make a new ImageDataGenerator with no augmentation?

In the example below, both `flow1, flow2` will have the same data, but no augmentation.

What do you think?

### Example
```python
import os

import numpy as np
from PIL import Image

pjoin = os.path.join

cls1_pt = 'tmp/cls1/cls1'
cls2_pt = 'tmp/cls1/cls2'
os.makedirs('tmp', exist_ok=True)
os.makedirs(cls1_pt, exist_ok=True)
os.makedirs(cls2_pt, exist_ok=True)

for i in range(100):
    img = Image.fromarray((np.ones([100, 100, 3]) * i % 255).astype(np.uint8))
    img.save(pjoin(cls1_pt, '{}.png'.format(i)))
    img.save(pjoin(cls2_pt, '{}.png'.format(i)))

flow1 = ImageDataGenerator(rotation_range=10, horizontal_flip=True).flow_from_directory('tmp/cls1',
                                                                                        seed=1337,
                                                                                        shuffle=False)
flow2 = ImageDataGenerator().flow_from_directory('tmp/cls1', seed=1337, shuffle=False)

assert flow1.filenames == flow2.filenames
```

cbernet:
Hi! no worries, really. 

In fact, we need a single ImageDataGenerator to be able to split the data between into the training and validation subsets.  For instance, in your example, all images appear in both of your flows, while we want to keep the training and validation subsets separated. 

In principle, it's possible to make the solution you propose work if the user: 

- disables shuffling, and sets the same seed for both generators as you did
- splits the two generators in the same way, say at 0.2
- uses the training subset of the first generator and the validation subset of the second one

I think that's very error prone and anyway, we need shuffling. 

From what I've seen, what people are doing at the moment is to split validation and training datasets physically on the disk, but that's very impractical. For instance, if you want to change the fraction of examples used for validation, you have to re-organize the data on the disk accordingly.

Dref360:
I'm pretty sure this is deterministic? Do you have repro steps? I agree that this should be better documented.

```python
import os

import numpy as np
from PIL import Image

pjoin = os.path.join

cls1_pt = 'tmp/cls1/cls1'
cls2_pt = 'tmp/cls1/cls2'
os.makedirs('tmp', exist_ok=True)
os.makedirs(cls1_pt, exist_ok=True)
os.makedirs(cls2_pt, exist_ok=True)

for i in range(100):
    img = Image.fromarray((np.ones([100, 100, 3]) * i % 255).astype(np.uint8))
    img.save(pjoin(cls1_pt, '{}.png'.format(i)))
    img.save(pjoin(cls2_pt, '{}.png'.format(i)))

for phase in ['training', 'validation']:
    flow1 = ImageDataGenerator(rescale=1 / 255.,
                               rotation_range=10, horizontal_flip=True,  validation_split=0.7).flow_from_directory('tmp/cls1', shuffle=True, subset=phase)
    flow2 = ImageDataGenerator(rescale=1 / 255.,  validation_split=0.7).flow_from_directory('tmp/cls1', shuffle=True, subset=phase)

    assert flow1.filenames == flow2.filenames

```

rragundez:
I agree with @Dref360 , what he proposes will achieved the desired functionality, and II think is more inline with the concept of what an `ImageDataGenerator` does. If you set augmentation, it will augment images, regardless.
Can we close this?

Dref360:
Thank for your PR, because there is an easy workaround we will close for now.

Issue #100, opened 2018-11-30
image rescaling and featurewise normalization not coordinated 


alexhernandezgarcia:
I believe the current implementation provides an erroneous, unexpected behaviour if both the `rescale` parameter is used (`not None` or different than 0) and feature-wise normalization is applied (`featurewise_center`, `featurewise_std_normalization`, ZCA whitening). 

 The `fit()` function computes the statistics from the original, _un-rescaled_ inputs and these statistics are applied finally on the _rescaled_ data. For instance, if the images are `uint8` (in the range [0, 255]) the feature-wise mean may be, for instance, 128. Then, if `rescale=1./255` the output images will be in the range [0, 1], but the original mean 128 will be subtracted.

```
import numpy as np
from keras.preprocessing.image import ImageDataGenerator

images = np.random.randint(low=0, high=255, size=(10, 32, 32, 3))

rescale = 1. / 255
imagedatagen = ImageDataGenerator(rescale=rescale,
                                  featurewise_center=True)
imagedatagen.fit(images)
batchgen = imagedatagen.flow(images, batch_size=10)
batch = batchgen.next()

images = images.astype(float)
images *= rescale
mean = np.mean(images, axis=(0, 1, 2))
images -= mean
print('Data range should be (approximately): [{}, {}]. \n'
      'Actual data range is: [{}, {}]'.format(np.min(images),
                                              np.max(images),
                                              np.min(batch),
                                              np.max(batch)))
```

PR #257, opened 2019-10-31
Rescale with featurewise normalization bug fix

hadipash:
### Summary
Featurewise centering, normalization and ZCA whitening work incorrectly with the `rescale` parameter. This PR fixes the bug.

### Related Issues
#100 

### PR Overview

### DIFF
--- a/keras_preprocessing/image/image_data_generator.py
+++ b/keras_preprocessing/image/image_data_generator.py
@@ -943,6 +943,9 @@ class ImageDataGenerator(object):
             np.random.seed(seed)
 
         x = np.copy(x)
+        if self.rescale:
+            x *= self.rescale
+
         if augment:
             ax = np.zeros(
                 tuple([rounds * x.shape[0]] + list(x.shape)[1:]),
--- a/keras_preprocessing/image/image_data_generator.py
+++ b/keras_preprocessing/image/image_data_generator.py
@@ -429,7 +429,8 @@ class ImageDataGenerator(object):
             save_to_dir=save_to_dir,
             save_prefix=save_prefix,
             save_format=save_format,
-            subset=subset
+            subset=subset,
+            dtype=self.dtype
         )
 
     def flow_from_directory(self,
@@ -537,7 +538,8 @@ class ImageDataGenerator(object):
             save_format=save_format,
             follow_links=follow_links,
             subset=subset,
-            interpolation=interpolation
+            interpolation=interpolation,
+            dtype=self.dtype
         )
 
     def flow_from_dataframe(self,
@@ -680,14 +682,15 @@ class ImageDataGenerator(object):
             save_format=save_format,
             subset=subset,
             interpolation=interpolation,
-            validate_filenames=validate_filenames
+            validate_filenames=validate_filenames,
+            dtype=self.dtype
         )
 
     def standardize(self, x):
         """Applies the normalization configuration in-place to a batch of inputs.
 
         `x` is changed in-place since the function is mainly used internally
-        to standarize images and feed them to your network. If a copy of `x`
+        to standardize images and feed them to your network. If a copy of `x`
         would be created instead it would have a significant performance cost.
         If you want to apply this method without changing the input in-place
         you can call the method creating a copy before:
@@ -909,6 +912,9 @@ class ImageDataGenerator(object):
 
         Only required if `featurewise_center` or
         `featurewise_std_normalization` or `zca_whitening` are set to True.
+
+        In case when `rescale` is set to a value, rescaling is applied to
+        sample data before computing the internal data stats.
 
         # Arguments
             x: Sample data. Should have rank 4.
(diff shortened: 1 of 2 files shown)


keras_preprocessing/image/image_data_generator.py
```diff
@@ -943,6 +943,9 @@ def fit(self, x,
             np.random.seed(seed)
 
         x = np.copy(x)
+        if self.rescale:
```
rragundez: Even though somewhat obvious can you add a comment explaining why rescaling before featurewise normalization, etc.. ?

tests/image/image_data_generator_test.py
```diff
@@ -442,5 +442,34 @@ def test_random_transforms():
...
+        images = np.vstack(img_list)
+
+        # featurewise_center and featurewise_std_normalization test
+        generator = image_data_generator.ImageDataGenerator(
+            rescale=rescale,
+            featurewise_center=True,
+            featurewise_std_normalization=True)
+        generator.fit(images)
+        batch = generator.flow(images, batch_size=8).next()
+        assert abs(np.mean(batch)) < 1e-3
```
rragundez: please make it a bit tighter 1E-6

tests/image/image_data_generator_test.py
```diff
@@ -442,5 +442,34 @@ def test_random_transforms():
...
+        assert abs(1 - np.std(batch)) < 1e-3
+
+        # zca_whitening test
+        generator = image_data_generator.ImageDataGenerator(
+            rescale=rescale,
+            featurewise_center=True,
+            zca_whitening=True)
+        generator.fit(images)
+        batch = generator.flow(images, batch_size=8).next()
+        assert np.max(np.abs(batch)) <= 1
```
rragundez: please check thee mean here as well.

keras_preprocessing/image/image_data_generator.py
hadipash: In the case of featurewise normalization and ZCA whitening rescaling isn’t really necessary since they apply their own kind of "scaling". So, this bug fix is needed more for consistent behavior no matter whether the `rescale` parameter is specified or not (before the code will produce unexpected results if the `rescale` parameter is specified).

The change to the code is useful when a user wants to apply rescaling with featurewise centering only. With this bug fix the code will produce correct outputs.

tests/image/image_data_generator_test.py
hadipash: Even without my changes to the code I couldn’t produce outputs with constant precision higher than 1e-6. Mostly it is in the range of 1e-4 to 1e-7 (tried on both Ubuntu and Windows).

keras_preprocessing/image/image_data_generator.py
rragundez: I meant a comment in the code, explaining to the developer why rescaling before...

tests/image/image_data_generator_test.py
rragundez: that's a bit weird, I would expect that precision level. Any ideas why or from where is the difference coming from?

hadipash: I found out why this happened. The reason is that even if I pass `dtype='float64'` to the `ImageDataGenerator`, it creates an iterator (`NumpyArrayIterator`, `DirectoryIterator`, or `DataFrameIterator`) using the default `dtype='float32'` thus decreasing the precision. I fixed this problem in the latest commit and set precision in tests to 1e-6.

tests/image/image_data_generator_test.py
hadipash: I digged deeper into the ZCA whitening topic and proposed a better test condition in the latest commit. Basically, after applying whitening, we obtain a new matrix **_Y_** (**_Y_** = **_WX_**, where **_W_** is a decorrelating matrix) where **_YY_**<sup>T</sup> is a diagonal matrix such that **_YY_**<sup>T</sup> = _(n-1)_**_I_** (_n_ is a number of data samples).

(More about this can be found in [Learning Multiple Layers of Features from Tiny Images](http://www.cs.toronto.edu/~kriz/learning-features-2009-TR.pdf) in Appendix A The ZCA whitening transformation)

hadipash:
Thanks @rragundez for your help! Few comments on the latest commit. I followed your suggestions on enhancing test cases but in the case of `featurewise_std_normalization` it's impossible to achieve 1e-6 precision because on the lines [722](https://github.com/keras-team/keras-preprocessing/blob/master/keras_preprocessing/image/image_data_generator.py#L722) and [967](https://github.com/keras-team/keras-preprocessing/blob/master/keras_preprocessing/image/image_data_generator.py#L967) of `image_data_generator.py` X is divided by the standard deviation + 1e-6 (`x /= (self.std + 1e-6)`), so the precision is smaller. Therefore, I set the level to 1e-5.

Additionally, I added `dtype` argument for iterators initialization so that data can be treated as other types rather than only `float32` (as explained above).

Finally, I introduced a better test case for ZCA whitening (also explained above).

keras_preprocessing/image/image_data_generator.py
```diff
@@ -910,6 +913,9 @@ def fit(self, x,
         Only required if `featurewise_center` or
         `featurewise_std_normalization` or `zca_whitening` are set to True.
 
+        In case when `rescale` is set to a value, rescaling is applied to
```
rragundez: ~In case~ When `rescale` ...

Issue #260, opened 2019-11-18
OSError: [Errno 24] Too many open files when using `load_img`


duhaime:
Please make sure that the boxes below are checked before you submit your issue.
If your issue is an **implementation question**, please ask your question on [StackOverflow](http://stackoverflow.com/questions/tagged/keras) or [on the Keras Slack channel](https://keras-slack-autojoin.herokuapp.com/) instead of opening a GitHub issue.

Thank you!

- [x] Check that you are up-to-date with the master branch of keras-preprocessing. You can update with:
`pip install git+git://github.com/keras-team/keras-preprocessing.git --upgrade --no-deps`

- [x] Provide a link to a GitHub Gist of a Python script that can reproduce your issue (or just copy the script here if it is short).

```python
from keras.preprocessing.image import load_img, save_img, array_to_img
import numpy as np

im = array_to_img(np.random.rand(300,300,3))
save_img('im.jpg', im)

# read in the image lots of times
# throws OSError: [Errno 24] Too many open files: 'im.jpg'
images = [load_img('im.jpg') for i in range(2**16)]
```

Smankusors:
Hmm wait, is this issue is still open or already solved by #261? Shouldn't this be closed?

duhaime:
Interesting, can you show the snippet that's causing this for you, and post your keras and/or keras_preprocessing versions? Here's a minimal snippet you could use for testing:

```
import numpy as np
from keras_preprocessing.image.utils import load_img, save_img, array_to_img
# or from keras.preprocessing.image import load_img, save_img, array_to_img

im = array_to_img(np.random.rand(300,300,3))
save_img('im.jpg', im)

# read in the image lots of times
# throws OSError: [Errno 24] Too many open files: 'im.jpg'
images = [load_img('im.jpg') for i in range(2**16)]
```

Smankusors:
oh no no no, I already tested your snippet (on Google Colab) and it's not throwing errors. Just wondering why this is not closed? 😅

Dref360:
Thank you I'll close!

PR #261, opened 2019-11-18
autoclose files opened in load_img

duhaime:
### Summary

The current `load_img` function opens files but never closes them, which can cause `OSError: [Errno 24] Too many open files` if one reads in too many images in a process.

### Related Issues

https://github.com/keras-team/keras-preprocessing/issues/260

### PR Overview

### DIFF
--- a/keras_preprocessing/image/utils.py
+++ b/keras_preprocessing/image/utils.py
@@ -134,7 +134,7 @@ def load_img(path, grayscale=False, color_mode='rgb', target_size=None,
                             ", ".join(_PIL_INTERPOLATION_METHODS.keys())))
                 resample = _PIL_INTERPOLATION_METHODS[interpolation]
                 img = img.resize(width_height_tuple, resample)
-    return img
+        return img
 
 
 def list_pictures(directory, ext=('jpg', 'jpeg', 'bmp', 'png', 'ppm', 'tif',
--- a/keras_preprocessing/image/utils.py
+++ b/keras_preprocessing/image/utils.py
@@ -4,6 +4,7 @@ from __future__ import absolute_import
 from __future__ import division
 from __future__ import print_function
 
+import io
 import os
 import warnings
 
@@ -109,7 +110,8 @@ def load_img(path, grayscale=False, color_mode='rgb', target_size=None,
     if pil_image is None:
         raise ImportError('Could not import PIL.Image. '
                           'The use of `load_img` requires PIL.')
-    with pil_image.open(path) as img:
+    with open(path, 'rb') as f:
+        img = pil_image.open(io.BytesIO(f.read()))
         if color_mode == 'grayscale':
             # if image is not already an 8-bit, 16-bit or 32-bit grayscale image
             # convert it to an 8-bit grayscale image.
--- a/keras_preprocessing/image/utils.py
+++ b/keras_preprocessing/image/utils.py
@@ -109,31 +109,31 @@ def load_img(path, grayscale=False, color_mode='rgb', target_size=None,
     if pil_image is None:
         raise ImportError('Could not import PIL.Image. '
                           'The use of `load_img` requires PIL.')
-    img = pil_image.open(path)
-    if color_mode == 'grayscale':
-        # if image is not already an 8-bit, 16-bit or 32-bit grayscale image
-        # convert it to an 8-bit grayscale image.
-        if img.mode not in ('L', 'I;16', 'I'):
-            img = img.convert('L')
-    elif color_mode == 'rgba':
-        if img.mode != 'RGBA':
-            img = img.convert('RGBA')
-    elif color_mode == 'rgb':
-        if img.mode != 'RGB':
-            img = img.convert('RGB')
-    else:
-        raise ValueError('color_mode must be "grayscale", "rgb", or "rgba"')
-    if target_size is not None:
-        width_height_tuple = (target_size[1], target_size[0])
-        if img.size != width_height_tuple:
-            if interpolation not in _PIL_INTERPOLATION_METHODS:
-                raise ValueError(
-                    'Invalid interpolation method {} specified. Supported '
-                    'methods are {}'.format(
-                        interpolation,
-                        ", ".join(_PIL_INTERPOLATION_METHODS.keys())))
-            resample = _PIL_INTERPOLATION_METHODS[interpolation]
-            img = img.resize(width_height_tuple, resample)
+    with pil_image.open(path) as img:
+        if color_mode == 'grayscale':
+            # if image is not already an 8-bit, 16-bit or 32-bit grayscale image
+            # convert it to an 8-bit grayscale image.
+            if img.mode not in ('L', 'I;16', 'I'):
+                img = img.convert('L')
+        elif color_mode == 'rgba':
+            if img.mode != 'RGBA':
+                img = img.convert('RGBA')
+        elif color_mode == 'rgb':
+            if img.mode != 'RGB':
+                img = img.convert('RGB')
+        else:
+            raise ValueError('color_mode must be "grayscale", "rgb", or "rgba"')
+        if target_size is not None:
+            width_height_tuple = (target_size[1], target_size[0])
+            if img.size != width_height_tuple:
+                if interpolation not in _PIL_INTERPOLATION_METHODS:
+                    raise ValueError(
+                        'Invalid interpolation method {} specified. Supported '
+                        'methods are {}'.format(
+                            interpolation,
+                            ", ".join(_PIL_INTERPOLATION_METHODS.keys())))
+                resample = _PIL_INTERPOLATION_METHODS[interpolation]
+                img = img.resize(width_height_tuple, resample)
     return img
 
 
--- a/keras_preprocessing/image/utils.py
+++ b/keras_preprocessing/image/utils.py
@@ -4,6 +4,7 @@ from __future__ import absolute_import
 from __future__ import division
 from __future__ import print_function
 
+import io
 import os
 import warnings
 
@@ -109,7 +110,8 @@ def load_img(path, grayscale=False, color_mode='rgb', target_size=None,
     if pil_image is None:
         raise ImportError('Could not import PIL.Image. '
                           'The use of `load_img` requires PIL.')
-    with pil_image.load(path) as img:
+    with open(path, 'rb') as f:
+        img = pil_image.open(io.BytesIO(f.read()))
         if color_mode == 'grayscale':
             # if image is not already an 8-bit, 16-bit or 32-bit grayscale image
             # convert it to an 8-bit grayscale image.
--- a/keras_preprocessing/image/utils.py
+++ b/keras_preprocessing/image/utils.py
@@ -4,7 +4,6 @@ from __future__ import absolute_import
 from __future__ import division
 from __future__ import print_function
 
-import io
 import os
 import warnings
 
@@ -110,8 +109,7 @@ def load_img(path, grayscale=False, color_mode='rgb', target_size=None,
     if pil_image is None:
         raise ImportError('Could not import PIL.Image. '
                           'The use of `load_img` requires PIL.')
-    with open(path, 'rb') as f:
-        img = pil_image.open(io.BytesIO(f.read()))
+    with pil_image.load(path) as img:
         if color_mode == 'grayscale':
             # if image is not already an 8-bit, 16-bit or 32-bit grayscale image
             # convert it to an 8-bit grayscale image.
--- a/keras_preprocessing/text.py
+++ b/keras_preprocessing/text.py
@@ -7,6 +7,7 @@ from __future__ import print_function
 
 import string
 import sys
+import six
 import warnings
 from collections import OrderedDict
 from collections import defaultdict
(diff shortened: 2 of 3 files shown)


keras_preprocessing/image/utils.py
```diff
@@ -109,32 +110,33 @@ def load_img(path, grayscale=False, color_mode='rgb', target_size=None,
     if pil_image is None:
         raise ImportError('Could not import PIL.Image. '
                           'The use of `load_img` requires PIL.')
-    img = pil_image.open(path)
```
rragundez: you can probably do `with pil_image.close(path) as img`, keeping everything in PILLOW land. Can you check?

duhaime:
Thanks for your quick followup @rragundez! I had thought the same, but it seems closing the filehandler before returning the object forbids a number of operations on the returned object. Minimal example:

```
import numpy as np
from PIL import Image as pil_image

im = pil_image.open('cats.jpg')
im.close()
im.getdata()
```

That will throw `ValueError: Operation on closed image`. 

For what it's worth, the approach here is straight from the [Pillow docs on file handling](https://pillow.readthedocs.io/en/4.0.x/reference/open_files.html). That said, I'm happy to make any changes you see fit!

keras_preprocessing/image/utils.py
duhaime: @rragundez Testing with `with pil_image.open(path) as img:`, I get the following errors:

<details>
 <summary>Error Report</summary>
(3.5) Duhaime:keras-preprocessing doug$ py.test tests/
================================================== test session starts ==================================================
platform darwin -- Python 3.5.3, pytest-3.0.6, py-1.8.0, pluggy-0.4.0
rootdir: /Users/doug/Desktop/keras-preprocessing, inifile:
plugins: pep8-1.0.6, cov-2.6.0, celery-4.1.1
collected 79 items

tests/sequence_test.py .........
tests/test_api.py ...
tests/test_documentation.py .
tests/text_test.py ....................
tests/image/affine_transformations_test.py ........
tests/image/dataframe_iterator_test.py ..............
tests/image/directory_iterator_test.py .....
tests/image/image_data_generator_test.py ..........
tests/image/iterator_test.py .
tests/image/numpy_array_iterator_test.py .
tests/image/test_image_api.py ..
tests/image/utils_test.py .F...

======================================================= FAILURES ========================================================
_____________________________________________________ test_load_img _____________________________________________________

tmpdir = local('/private/var/folders/cy/ywx04ttn12n_xmgjf5s6sbzr0000gn/T/pytest-of-doug/pytest-49/test_load_img0')

    def test_load_img(tmpdir):
        filename_rgb = str(tmpdir / 'rgb_utils.png')
        filename_rgba = str(tmpdir / 'rgba_utils.png')
        filename_grayscale_8bit = str(tmpdir / 'grayscale_8bit_utils.png')
        filename_grayscale_16bit = str(tmpdir / 'grayscale_16bit_utils.tiff')
        filename_grayscale_32bit = str(tmpdir / 'grayscale_32bit_utils.tiff')

        original_rgb_array = np.array(255 * np.random.rand(100, 100, 3),
                                      dtype=np.uint8)
        original_rgb = utils.array_to_img(original_rgb_array, scale=False)
        original_rgb.save(filename_rgb)

        original_rgba_array = np.array(255 * np.random.rand(100, 100, 4),
                                       dtype=np.uint8)
        original_rgba = utils.array_to_img(original_rgba_array, scale=False)
        original_rgba.save(filename_rgba)

        original_grayscale_8bit_array = np.array(255 * np.random.rand(100, 100, 1),
                                                 dtype=np.uint8)
        original_grayscale_8bit = utils.array_to_img(original_grayscale_8bit_array,
                                                     scale=False)
        original_grayscale_8bit.save(filename_grayscale_8bit)

        original_grayscale_16bit_array = np.array(
            np.random.randint(-2147483648, 2147483647, (100, 100, 1)), dtype=np.int16
        )
        original_grayscale_16bit = utils.array_to_img(original_grayscale_16bit_array,
                                                      scale=False, dtype='int16')
        original_grayscale_16bit.save(filename_grayscale_16bit)

        original_grayscale_32bit_array = np.array(
            np.random.randint(-2147483648, 2147483647, (100, 100, 1)), dtype=np.int32
        )
        original_grayscale_32bit = utils.array_to_img(original_grayscale_32bit_array,
                                                      scale=False, dtype='int32')
        original_grayscale_32bit.save(filename_grayscale_32bit)

        # Test that loaded image is exactly equal to original.

        loaded_im = utils.load_img(filename_rgb)
>       loaded_im_array = utils.img_to_array(loaded_im)

tests/image/utils_test.py:61:
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _
/Users/doug/anaconda/envs/3.5/lib/python3.5/site-packages/Keras_Preprocessing-1.1.0-py3.5.egg/keras_preprocessing/image/utils.py:309: in img_to_array
    x = np.asarray(img, dtype=dtype)
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _

a = <PIL.PngImagePlugin.PngImageFile image mode=RGB size=100x100 at 0x127774208>, dtype = 'float32', order = None

    @set_module('numpy')
    def asarray(a, dtype=None, order=None):
        """Convert the input to an array.

        Parameters
        ----------
        a : array_like
            Input data, in any form that can be converted to an array.  This
            includes lists, lists of tuples, tuples, tuples of tuples, tuples
            of lists and ndarrays.
        dtype : data-type, optional
            By default, the data-type is inferred from the input data.
        order : {'C', 'F'}, optional
            Whether to use row-major (C-style) or
            column-major (Fortran-style) memory representation.
            Defaults to 'C'.

        Returns
        -------
        out : ndarray
            Array interpretation of `a`.  No copy is performed if the input
            is already an ndarray with matching dtype and order.  If `a` is a
            subclass of ndarray, a base class ndarray is returned.

        See Also
        --------
        asanyarray : Similar function which passes through subclasses.
        ascontiguousarray : Convert input to a contiguous array.
        asfarray : Convert input to a floating point ndarray.
        asfortranarray : Convert input to an ndarray with column-major
                         memory order.
        asarray_chkfinite : Similar function which checks input for NaNs and Infs.
        fromiter : Create an array from an iterator.
        fromfunction : Construct an array by executing a function on grid
                       positions.

        Examples
        --------
        Convert a list into an array:

        >>> a = [1, 2]
        >>> np.asarray(a)
        array([1, 2])

        Existing arrays are not copied:

        >>> a = np.array([1, 2])
        >>> np.asarray(a) is a
        True

        If `dtype` is set, array is copied only if dtype does not match:

        >>> a = np.array([1, 2], dtype=np.float32)
        >>> np.asarray(a, dtype=np.float32) is a
        True
        >>> np.asarray(a, dtype=np.float64) is a
        False

        Contrary to `asanyarray`, ndarray subclasses are not passed through:

        >>> issubclass(np.recarray, np.ndarray)
        True
        >>> a = np.array([(1.0, 2), (3.0, 4)], dtype='f4,i4').view(np.recarray)
        >>> np.asarray(a) is a
        False
        >>> np.asanyarray(a) is a
        True

        """
>       return array(a, dtype, copy=False, order=order)
E       TypeError: float() argument must be a string or a number, not 'PngImageFile'

../../anaconda/envs/3.5/lib/python3.5/site-packages/numpy/core/numeric.py:538: TypeError
========================================= 1 failed, 78 passed in 12.10 seconds ==========================================
</details>

I'm happy to try other potential solutions, but just wanted to mention that the method in this pull request is the [only self-closing method endorsed by the Pillow docs](https://pillow.readthedocs.io/en/4.0.x/reference/open_files.html).

rragundez:
Thanks for your explanations. After having a closer look I think the root of the problem is the inconsistency `load_img` can return only a pointer to the file if no data augmentation is done (keeps file open) and an actual image in memory if data augmentation is done. I think this inconsistency is gone by simply changing `img = pil_image.open(path)` to `img = pil_image.load(path)` in
https://github.com/keras-team/keras-preprocessing/blob/master/keras_preprocessing/image/utils.py#L112

I would do that change and simply add a unit-test that `load_img` is actually returning an image in memory and not a pointer to the file, with and without augmentation.

duhaime:
@rragundez  Your approach (with pil_image.load(path) as img) is nice and clean, nice research. I just replaced the `io.BytesIO` business with that method, and added a test to validate that `load_img` does indeed return an image in memory.

How does the updated PR look?

duhaime:
@rragundez ah I was testing against a cached version of the module--it appears the load method is missing from PIL.Image--where did you see that method?

rragundez:
https://pillow.readthedocs.io/en/3.1.x/reference/Image.html?highlight=load#PIL.Image.Image.load

duhaime:
Hmm, these two return different objects:
```
pil_image.open('a.png')        # returns `PIL.PngImagePlugin.PngImageFile` object
pil_image.open('a.png').load() # returns `PixelAccess` object 
```

I understand the [Pillow docs on the `open` method](https://pillow.readthedocs.io/en/3.1.x/reference/Image.html#PIL.Image.open) to indicate that `PIL.Image.open()` opens the file (and never closes it) but waits to read the pixel data into RAM. `load()` evidently is a method that one can call to manually force the `PIL.Image` object to yield up its pixel data.

That said, combining open and load don't seem to close the file handlers. Returning PixelAccess data rather than a PIL Image object also breaks some methods in `keras_preprocessing`; e.g. you can't call `img_to_array` on the PixelAccess data:

```
>>> from keras.preprocessing.image import img_to_array
Using TensorFlow backend.
>>> from PIL import Image as pil_image
>>> a = pil_image.open('a.png').load()
>>> img_to_array(a)
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/Users/doug/anaconda/envs/3.5/lib/python3.5/site-packages/keras/preprocessing/image.py", line 75, in img_to_array
    return image.img_to_array(img, data_format=data_format, dtype=dtype)
  File "/Users/doug/anaconda/envs/3.5/lib/python3.5/site-packages/keras_preprocessing/image/utils.py", line 299, in img_to_array
    x = np.asarray(img, dtype=dtype)
  File "/Users/doug/anaconda/envs/3.5/lib/python3.5/site-packages/numpy/core/numeric.py", line 538, in asarray
    return array(a, dtype, copy=False, order=order)
TypeError: float() argument must be a string or a number, not 'PixelAccess'
```

It's entirely possible I misunderstood you or am missing something! From my perspective, using the io stream method outlined in the Pillow docs seems the best way forward...

keras_preprocessing/text.py
```diff
@@ -43,8 +44,10 @@ def text_to_word_sequence(text,
         text = text.lower()
 
     if sys.version_info < (3,):
-        if isinstance(text, unicode):
-            translate_map = dict((ord(c), unicode(split)) for c in filters)
+        # in Python 2.x six.string_types == basestring
```
rragundez: can you put this in a separate PR? thanks

tests/image/utils_test.py
```diff
@@ -326,5 +327,26 @@ def test_array_to_img_and_img_to_array():
...
+    path = write_sample_image(tmpdir)
+    max_open_files, _ = soft, hard = resource.getrlimit(resource.RLIMIT_NOFILE)
+    for i in range(max_open_files+1):
+        utils.load_img(path)
+
+
+def test_load_img_returns_image(tmpdir):
+    path = write_sample_image(tmpdir)
+    im = utils.load_img(path)
+    assert hasattr(im, 'height')
```
rragundez: can you check the type of the object instead? thanks

tests/image/utils_test.py
```diff
@@ -326,5 +327,26 @@ def test_array_to_img_and_img_to_array():
...
+def write_sample_image(tmpdir):
+    im = utils.array_to_img(np.random.rand(1, 1, 3))
+    path = str(tmpdir / 'sample_image.png')
+    utils.save_img(path, im)
+    return path
+
+
+def test_image_file_handlers_close(tmpdir):
+    path = write_sample_image(tmpdir)
+    max_open_files, _ = soft, hard = resource.getrlimit(resource.RLIMIT_NOFILE)
```
rragundez: Please keep a single assignment.

duhaime:
@rragundez sounds good! I just submitted some updates with the changes you requested. The flake8 checks are of course going to fail because of the undefined `unicode` in the text module. I factored the changes that rectify that problem into [a separate PR](https://github.com/keras-team/keras-preprocessing/pull/264) per your request above.

rragundez:
I think is good now, I'll have a closer look tomorrow :)

keras_preprocessing/text.py
```diff
@@ -514,4 +514,4 @@ def tokenizer_from_json(json_string):
     tokenizer.word_index = word_index
     tokenizer.index_word = index_word
 
-    return tokenizer
```
rragundez: Please remove any changes to this file.

duhaime: sure thing (my IDE auto-trims trailing whitespace on file save but I'll restore it)

rragundez: that's the reason the CI is faailing https://travis-ci.org/keras-team/keras-preprocessing/jobs/617265242?utm_medium=notification&utm_source=github_status

duhaime: ah, I see! I'll use vim instead

duhaime:
thanks for your help with this!

rragundez:
No problem, I don't understand why I still see changes in `keras_preprocessing/text.py`

rragundez:
thanks for the work @duhaime!

PR #270, opened 2019-12-31
To solve issue #13637 in keras repo

seriousran:
### Summary

Set function sort 'classes'
To solve it, use OrderedDict

### Related Issues

https://github.com/keras-team/keras/issues/13637

### PR Overview

### DIFF
--- a/keras_preprocessing/image/dataframe_iterator.py
+++ b/keras_preprocessing/image/dataframe_iterator.py
@@ -241,7 +241,8 @@ class DataFrameIterator(BatchFromFilesMixin, Iterator):
                 )
 
         if classes:
-            classes = list(OrderedDict.fromkeys(classes).keys()) # prepare for membership lookup
+            # prepare for membership lookup
+            classes = list(OrderedDict.fromkeys(classes).keys())
             df[y_col] = df[y_col].apply(lambda x: remove_classes(x, classes))
         else:
             classes = set()
--- a/keras_preprocessing/image/dataframe_iterator.py
+++ b/keras_preprocessing/image/dataframe_iterator.py
@@ -8,6 +8,7 @@ import os
 import warnings
 
 import numpy as np
+from collections import OrderedDict
 
 from .iterator import BatchFromFilesMixin, Iterator
 from .utils import validate_filename
@@ -240,7 +241,7 @@ class DataFrameIterator(BatchFromFilesMixin, Iterator):
                 )
 
         if classes:
-            classes = set(classes)  # sort and prepare for membership lookup
+            classes = list(OrderedDict.fromkeys(classes).keys()) # prepare for membership lookup
             df[y_col] = df[y_col].apply(lambda x: remove_classes(x, classes))
         else:
             classes = set()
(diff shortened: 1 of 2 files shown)


rragundez:
The parameter `classes` is for indicate which classes to take into account. So even if the `classes` are given in a different order by the user the `class_indices` maping will not change which the desired functionality given the `classes` parameter definition:  "Optional list of strings, classes to use (e.g. `["dogs", "cats"]`)."

seriousran:
@rragundez But, flow_from_directory and flow_from_dataframe work differently as [keras-team/keras#13637](https://github.com/keras-team/keras/issues/13637)

rragundez:
agree, that is a different issue than "not working" though. Let me check with @Dref360 . What do you think? I believe it should be one way that `classes` should work. On one hand I think it should independent on the order the user gives (serving just as a filter and nothing more, this also allows to give as a set), but on the other `flow_from_directory` predates by far `flow_from_dataframe` so perhaps for backwards compatibility is better to use hte `flow_from_directory` behaviour.

Dref360:
I agree with you that *_directory is far more used and we should stick to its behavior.

I think this PR fixes the behavior?
If it is the case, please add a test following keras-team/keras#13637 repro steps and edit the documentation accordingly.

qinst64:
@rragundez This is a bug as my exmaple. The class order shouldn't be reset silently, this leads to unknown class order of predicted results, and  then unknown calcluation of confustion matrix, and precition, recall etc (i.e. which class is definded as positive?)
Again with my example, if `flow_from_directory` (right behavior) then
  -   `classes=['A', 'B']` defines A=0 and B=1,
  -  `classes=['B', 'A']` defines B=0 and A=1
Otherwise say if predicted result = [[0.1, 0.2]], how would it be tell A=0.1 or B=0.1?

`flow_from_directory` has the right behavior for users to set which is the positive class; `flow_from_dataframe` doesn't.

rragundez:
@qinst64 `class_indices` gives the order.

@seriousran I don't really like the behaviour but as we point out is better to have a single one and the `*_directory` is used far more. I'll try to check the PR this weekend. Please add tests thanks!

rragundez:
Please add unit tests, I cannot review nor merge this PR without them. Thanks!

seriousran:
Couldn't pass the test with this error. But I can't understand why the rror occured in directory_iterator.py file. my commits didn't include that file. I will check it out again.

tests/image/directory_iterator_test.py ..*** Error in `/home/travis/miniconda/envs/test-environment/bin/python': double free or corruption (!prev): 0x0000564167c7d9c0 ***
Fatal Python error: Aborted

Thread 0x00007f253a71b740 (most recent call first):
  File "/home/travis/build/keras-team/keras-preprocessing/keras_preprocessing/image/directory_iterator.py", line 140 in __init__
  File "/home/travis/build/keras-team/keras-preprocessing/keras_preprocessing/image/image_data_generator.py", line 542 in flow_from_directory
  File "/home/travis/build/keras-team/keras-preprocessing/tests/image/directory_iterator_test.py", line 204 in test_directory_iterator_with_validation_split
...

seriousran:
I gave up. In local, it works well. All pytest passed. I can't handle this problem.

However, the error can be solved by this.
https://github.com/tensorflow/tensorflow/issues/8717#issuecomment-289828646

Dref360:
You need to remove `dist: trusty` in travis.yml

seriousran:
@Dref360 thanks :)

seriousran:
@rragundez 

I would appreciate it if you reconsider once again.

keras-team/keras#13637

PR #300, opened 2020-06-08
Add scale param to brightness and fix some docs

### LINKED ISSUE #114: random_brightness shifts the image normalization
My images are trained with float representation, so that their maximum value is 1.0.
However, when I applied `random_brightness`, the image is between 0.0 and 255.0

I think this is not expected, or at least, it should be warned in the documentation, shouldn't it?

Smankusors:
### Summary
This add an option to not scale the image (to 0 - 255) when doing `random_brightness`. This also might fixes #114.

Code example :
```
brightness_range = [0.5, 2]
image = random_brightness(image, brightness_range, True) #for enable scale
image = random_brightness(image, brightness_range, False) #for disable scale

#or alternatively
image = random_brightness(image, brightness_range, scale=True)
image = random_brightness(image, brightness_range, scale=False)

#or call the apply_brightness_shift directly
image = apply_brightness_shift(image, 0.5, scale=True)
```

### Related Issues
There's no documentation about this on keras.io...

### PR Overview

- [x] This PR requires new unit tests [n] (make sure tests are included)
- [ ] This PR requires to update the documentation [?] (make sure the docs are up-to-date)
- [x] This PR is backwards compatible [y]
- [x] This PR changes the current API [n] (all API changes need to be approved by fchollet)

### DIFF
--- a/keras_preprocessing/image/affine_transformations.py
+++ b/keras_preprocessing/image/affine_transformations.py
@@ -195,37 +195,41 @@ def random_channel_shift(x, intensity_range, channel_axis=0):
     return apply_channel_shift(x, intensity, channel_axis=channel_axis)
 
 
-def apply_brightness_shift(x, brightness):
+def apply_brightness_shift(x, brightness, scale=True):
     """Performs a brightness shift.
 
     # Arguments
         x: Input tensor. Must be 3D.
         brightness: Float. The new brightness value.
-        channel_axis: Index of axis for channels in the input tensor.
+        scale: Whether to rescale the image such that minimum and maximum values
+            are 0 and 255 respectively.
+            Default: True.
 
     # Returns
         Numpy image tensor.
 
     # Raises
-        ValueError if `brightness_range` isn't a tuple.
+        ImportError: if PIL is not available.
     """
     if ImageEnhance is None:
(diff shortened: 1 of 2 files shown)


Dref360:
Hello, thanks for the PR,
Could you add some tests that check that the new behavior is respected?

Smankusors:
> Hello, thanks for the PR,
> Could you add some tests that check that the new behavior is respected?

hmm currently I'm on Windows. But I have trouble running tests. I got 

```
E   ModuleNotFoundError: No module named 'resource'
```

I think I will need to use Linux to run tests? I already tried `pip3 install resource` but still module not found 😢

Smankusors:
anyway, what do you think about this?

```
def test_random_brightness_scale():
    img = np.ones((1, 1, 3)) * 128
    zeros = np.zeros((1, 1, 3))
    assert np.array_equal(img, affine_transformations.random_brightness(img, [1, 1], scale=False))
    assert np.array_equal(zeros, affine_transformations.random_brightness(img, [1, 1], scale=True))
```

Dref360:
Yeah looks good.
Yup you need UNIX to run the rest. We should probably change that.

Smankusors:
alright, I run the tests on the Google Colab 🤣 

And I made little changes so the tests can run successfully.

PR #303, opened 2020-07-15
Updated keras_preprocessing image utils

novasush:
### Summary
I have made some changes, that can be useful for people working on image processing.

1. image.load_img()
Gave me TypeError for loading bytes image, corrected the function to read an image from the path and BytesIO object.

2. image.array_to_img()
Used to fail on grayscale images because their ndim was less than 3, now added an optional flag is_grayscale to add single channel to grayscale image array with ndim = 2

3. image.img_to_array()
Replaced np.asarray() with np.array(), can be useful with cv2 functions, earlier np.asarray processed images gave different output when used with cv2 functions because array used to be modified. Now a copy of array is processed for use with cv2
### PR Overview

- [x] This PR changes the current API [y/n] (all API changes need to be approved by fchollet)

### DIFF
--- a/PULL_REQUEST_TEMPLATE.md
+++ b/PULL_REQUEST_TEMPLATE.md
@@ -4,7 +4,7 @@
 
 ### PR Overview
 
-- [n] This PR requires new unit tests [y/n] (make sure tests are included)
-- [y] This PR requires to update the documentation [y/n] (make sure the docs are up-to-date)
-- [y] This PR is backwards compatible [y/n]
-- [n] This PR changes the current API [y/n] (all API changes need to be approved by fchollet)
+- [x] This PR requires new unit tests [y/n] (make sure tests are included)
+- [x] This PR requires to update the documentation [y/n] (make sure the docs are up-to-date)
+- [x] This PR is backwards compatible [y/n]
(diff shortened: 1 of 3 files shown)


PULL_REQUEST_TEMPLATE.md
```diff
@@ -4,7 +4,7 @@
 
 ### PR Overview
 
-- [ ] This PR requires new unit tests [y/n] (make sure tests are included)
-- [ ] This PR requires to update the documentation [y/n] (make sure the docs are up-to-date)
-- [ ] This PR is backwards compatible [y/n]
-- [ ] This PR changes the current API [y/n] (all API changes need to be approved by fchollet)
+- [n] This PR requires new unit tests [y/n] (make sure tests are included)
```
Dref360, line 7: Please revert, those should be markdown checkboxes
[x] or [ ]

novasush: ok got it

Dref360:
I don't think reverting this [commit](https://github.com/keras-team/keras-preprocessing/commit/9a836c25177e1be5940e1b2ab19fdb383225c32a#diff-d16be738737f6fddf4ab05a2d02ef21a) is the right way to fix this issue, could you look into it?

PR #309, opened 2020-10-08
Add custom transformations

DOLARIK:
### Summary 
This PR will help us apply custom transformations to the images at the time of data augmentation.

Introduces a new argument, __list_of_custom_transformations__, in `ImageDataGenerator`, wherein we can pass a list of all the custom transformations that we want to apply to the images.

__For Example:__

Consider the following custom transformations:
```python3
# Custom Transformation 1
def transformation_one(x):
  """
  transformation 1 code here.
  """
  return transformed_x

# Custom Transformation 2
def transformation_two(x):
  """
  transformation 2 code here.
  """
  return transformed_x

# Custom Transformation 3
class TransformationThree:
   ...

   def apply_transform(self, x):
      """
      transformation 3 code here.
      """
      return transformed_x

transformation_three = TransformationThree()
```
Now the above two custom transformations are to be passed to __`list_of_custom_transformations`__  while initialising __`ImageDataGenerator`__ as shown below:

```python3

datagen = ImageDataGenerator(rotation_range = 90,
           zoom_range = 0.1, 
           shear_range = 0.3,
           width_shift_range=0.08,   
           height_shift_range=0.08,   
           vertical_flip=True,
           # custom transforms to be passed in list
           list_of_custom_transformations = [transformation_one, 
                                             transformation_two,
                                             transformation_three.apply_transform] 
)
```
And then the subsequent steps are as usual:

```python
# train_gen
train_gen = datagen.flow(train_images, train_labels, batch_size = 32)
```

I believe this feature will help us during the research phase of a project for trying out new and custom transformations.

I have __not__ added new tests for this feature yet. So, for you to test this out I have created this [Colab Notebook](https://colab.research.google.com/drive/16Fx6IfX07k4ECwF7qef3khR-lPmKGO-o?usp=sharing). __It includes demo transformations and visualizations.__

This is __my first PR__ and I am still figuring out the best practices for adding new tests and updating the docs. So, it would be __very helpful if someone can share some useful links to articles on creating new unit tests.__

Also, I didn't understand the 3rd point in the __PR Overview__ below. So, pls help me out with that too :smile: 

### PR Overview

- [x] This PR requires new unit tests [y] (have not included yet)
- [x] This PR requires to update the documentation [y] (have not updated yet)
- [ ] This PR is backwards compatible [?]
- [x] This PR changes the current API [y] (all API changes need to be approved by fchollet)

### DIFF
--- a/keras_preprocessing/image/image_data_generator.py
+++ b/keras_preprocessing/image/image_data_generator.py
@@ -102,6 +102,7 @@ class ImageDataGenerator(object):
             (strictly between 0 and 1).
         interpolation_order: int, order to use for
             the spline interpolation. Higher is slower.
+        list_of_custom_transformations: list, list of transformation functions
         dtype: Dtype to use for the generated arrays.
 
     # Examples
@@ -274,6 +275,7 @@ class ImageDataGenerator(object):
                  data_format='channels_last',
                  validation_split=0.0,
                  interpolation_order=1,
+                 list_of_custom_transformations=None,
                  dtype='float32'):
 
         self.featurewise_center = featurewise_center
@@ -296,6 +298,7 @@ class ImageDataGenerator(object):
         self.preprocessing_function = preprocessing_function
         self.dtype = dtype
         self.interpolation_order = interpolation_order
+        self.list_of_custom_transformations = list_of_custom_transformations
 
         if data_format not in {'channels_last', 'channels_first'}:
             raise ValueError(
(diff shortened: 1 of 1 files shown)


Dref360:
Hello!
Thank you for your PR.
I am not sure if this is the right way to achieve what you want.

Recently, the TF team released [Keras Preprocessing Layers](https://keras.io/guides/preprocessing_layers/) which would apply transformations after the data augmentation (to add further transformations for example).

Is there a use-case where these layers are not working for you? If so, could you describe them so that we can fix that?

PR #314, opened 2020-10-28
Set matrix center correctly

eli-osherovich:
Matrix center was set incorrectly (one off). To correct either call transform_matrix_offset_center() with h-1 ,w-1. Or  fix the sign inside that function and continue using h and w (more intuitive).

### PR Overview

### DIFF
--- a/keras_preprocessing/image/affine_transformations.py
+++ b/keras_preprocessing/image/affine_transformations.py
@@ -247,8 +247,8 @@ def random_brightness(x, brightness_range, scale=True):
 
 
 def transform_matrix_offset_center(matrix, x, y):
-    o_x = float(x) / 2 + 0.5
-    o_y = float(y) / 2 + 0.5
+    o_x = float(x) / 2 - 0.5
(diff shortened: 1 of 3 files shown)


Dref360:
Could we add a test to be sure that the matrix is rightly centered?

eli-osherovich:
@Dref360 Do you mean a unit test or the way to know wither + or - is correct?

Here is a simple example:
```python
img = np.expand_dims(np.array([[0, 1], [0, 0]]), -1)
tf.keras.preprocessing.image.apply_affine_transform(img, theta=90)

array([[[0],
        [0]],

       [[0],
        [0]]])
```

Dref360:
Yes so that if we modify this portion of the code again, we won't merge the same bug.

eli-osherovich:
Added a unit test.

P. S. 
There are two existing tests that seem to fail. 
These should be fixed in a separate PR.

Dref360 (approved):
LGTM

PR #324, opened 2020-11-01
Fixes #323.

### LINKED ISSUE #323: Bug: axes transformation is incomplete
Please make sure that the boxes below are checked before you submit your issue.
If your issue is an **implementation question**, please ask your question on [StackOverflow](http://stackoverflow.com/questions/tagged/keras) or [on the Keras Slack channel](https://keras-slack-autojoin.herokuapp.com/) instead of opening a GitHub issue.

Thank you!

- [x] Check that you are up-to-date with the master branch of keras-preprocessing. You can update with:
`pip install git+git://github.com/keras-team/keras-preprocessing.git --upgrade --no-deps`

- [x] Provide a link to a GitHub Gist of a Python script that can reproduce your issue (or just copy the script here if it is short).

I introduced this bug in a recent PR that exchanged the X and Y axes. 
Forgot final transformation back to the original data order. Hence, the image is transposed.

eli-osherovich:
### PR Overview

Fixes #323

### DIFF
--- a/tests/image/affine_transformations_test.py
+++ b/tests/image/affine_transformations_test.py
@@ -39,8 +39,8 @@ def test_matrix_center():
         [0, 0],
     ]), -1)
     x_rotated90 = np.expand_dims(np.array([
+        [1, 0],
         [0, 0],
-        [0, 1],
     ]), -1)
 
     assert np.allclose(
--- a/keras_preprocessing/image/affine_transformations.py
+++ b/keras_preprocessing/image/affine_transformations.py
@@ -372,9 +372,16 @@ def apply_affine_transform(x, theta=0, tx=0, ty=0, shear=0, zx=1, zy=1,
 
         # Matrix construction assumes that coordinates are x, y (in that order).
         # However, regular numpy arrays use y,x (aka i,j) indexing.
-        # Swap the first two columns of the matrix to simulate the x, y indexing.
+        # Possible solution is:
+        #   1. Swap the x and y axes.
+        #   2. Apply transform.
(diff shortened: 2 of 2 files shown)


tests/image/affine_transformations_test.py
```diff
@@ -39,8 +39,8 @@ def test_matrix_center():
         [0, 0],
     ]), -1)
     x_rotated90 = np.expand_dims(np.array([
+        [1, 0],
```
Dref360, line 42: The rotation should be clock-wise I think.

eli-osherovich: It happened so that I re-wrote the whole module yesterday and spent some time figuring out exact directions and quantities of each transform. I concluded that the rotation should be CCW, which led me to finding this bug. 

The reasoning is as follows:
1. The Y axis in images points down, hence the positive rotation should be CW. Consequently, the negative direction is CCW.
2. Affine transform used here is the inverse (aka backward, pull, etc.)  transform (see [affine_transform](https://docs.scipy.org/doc/scipy/reference/generated/scipy.ndimage.affine_transform.html)). Therefore, it "negates" its arguments. Hence, for positive angle, the rotation is negative, namely, CCW. 
Similarly, for positive translation, the image moves in the negative direction. 

P. S.
As an additional sanity check,  I verified that the behavior here matches [scikit-image](https://scikit-image.org/docs/dev/api/skimage.transform.html#skimage.transform.AffineTransform)

Dref360 (approved):

PR #333, opened 2021-01-19
Keep original aspect ratio of objects in flow_from_directory

kopytjuk:
### Summary

Often while using `ImageDataGenerator` from local images, you want to maintain the aspect ratio of the objects itself in order to avoid models to learn the retransformation. This flag allows to maintain original aspect ratio of the objects.

An additional parameter in `flow_from_directory` is introduced, i.e. `train_datagen.flow_from_directory(keep_aspect_ratio=True)`.

before

after

In `tensorflow` there is https://www.tensorflow.org/api_docs/python/tf/keras/preprocessing/image/smart_resize?hl=en , copied the logic from there, since reuse was not possible due circular import ...

### Related Issues

https://github.com/keras-team/keras/pull/4987/files

https://stackoverflow.com/questions/42467734/keras-how-to-use-imagedatagenerator-without-deforming-aspect-ratio

### PR Overview

- [y] This PR changes the current API [y/n] (all API changes need to be approved by fchollet)

### DIFF
--- a/keras_preprocessing/image/directory_iterator.py
+++ b/keras_preprocessing/image/directory_iterator.py
@@ -60,7 +60,7 @@ class DirectoryIterator(BatchFromFilesMixin, Iterator):
             If PIL version 1.1.3 or newer is installed, "lanczos" is also
             supported. If PIL version 3.4.0 or newer is installed, "box" and
             "hamming" are also supported. By default, "nearest" is used.
-        keep_aspect_ratio: Boolean, whether to resize images to a target size 
+        keep_aspect_ratio: Boolean, whether to resize images to a target size
             without aspect ratio distortion. The image is cropped in the center
             with target aspect ratio before resizing.
         dtype: Dtype to use for generated arrays.
--- a/keras_preprocessing/image/directory_iterator.py
+++ b/keras_preprocessing/image/directory_iterator.py
@@ -68,6 +68,7 @@ class DirectoryIterator(BatchFromFilesMixin, Iterator):
                  directory,
                  image_data_generator,
                  target_size=(256, 256),
+                 keep_aspect_ratio=False,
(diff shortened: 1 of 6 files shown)


keras_preprocessing/image/utils.py
```diff
@@ -77,7 +77,7 @@ def save_img(path,
 
 
 def load_img(path, grayscale=False, color_mode='rgb', target_size=None,
-             interpolation='nearest'):
+             interpolation='nearest', keep_aspect_ratio=False):
     """Loads an image into PIL format.
```
Dref360, line 81: Please add documentation

keras_preprocessing/image/image_data_generator.py
```diff
@@ -463,6 +464,8 @@ def flow_from_directory(self,
             target_size: Tuple of integers `(height, width)`,
                 default: `(256, 256)`.
                 The dimensions to which all images found will be resized.
+            keep_aspect_ratio:
```
Dref360, line 467: Please add documentation

keras_preprocessing/image/directory_iterator.py
```diff
@@ -68,6 +68,7 @@ def __init__(self,
                  directory,
                  image_data_generator,
                  target_size=(256, 256),
+                 keep_aspect_ratio=False,
```
Dref360, line 71: Please put the new argument toward the end of the constructor. After `interpolation` would be good.

Dref360 (changes requested):
Hello, I think we can merge this once the following are done:
* Put the argument near `interpolation` in the ImageDataGenerator, otherwise we will cause backward compatibility issues .
* Add some tests that the behavior is correct.
* Add documentation.

Dref360 (approved):
LGTM thanks!

Issue #322, opened 2020-11-01
Get rid of scipy


eli-osherovich:
Please make sure that the boxes below are checked before you submit your issue.
If your issue is an **implementation question**, please ask your question on [StackOverflow](http://stackoverflow.com/questions/tagged/keras) or [on the Keras Slack channel](https://keras-slack-autojoin.herokuapp.com/) instead of opening a GitHub issue.

Thank you!

- [x] Check that you are up-to-date with the master branch of keras-preprocessing. You can update with:
`pip install git+git://github.com/keras-team/keras-preprocessing.git --upgrade --no-deps`

- [x] Provide a link to a GitHub Gist of a Python script that can reproduce your issue (or just copy the script here if it is short).

From my experience, relying on SciPy for image transformation makes it:
a) slow
b) non-suitable for TPU

Is there any particular reason to keep SciPy in the loop?

Stanfording:
For the m1 apple chip, SciPy is not available yet, but the ImageDataGenerator.flow_from_directory requires it. So getting rid of scipy would be a temporary relief for the m1 users. Maybe not necessary those since SciPy is working on it too I guess.

eli-osherovich:
I can do this if there is a real need from the keras side.

Dref360:
We use scipy here for SVD (I think np.linalg.svd is called?)
[here](https://github.com/keras-team/keras-preprocessing/blob/54b5d6ed65f1608ca7207d585075c2ad3a52929e/keras_preprocessing/image/image_data_generator.py#L986)

And we use scipy.ndimage for matrix multiplication here:
[ndimage.interpolation.affine_transform](https://github.com/keras-team/keras-preprocessing/blob/54b5d6ed65f1608ca7207d585075c2ad3a52929e/keras_preprocessing/image/affine_transformations.py#L387). 

Do you think it would require a lot of work to replicate `affine_transform` behaviour?

eli-osherovich:
No. It should be pretty simple.

Dref360:
PRs welcome then :)

eli-osherovich:
Will send some shortly

Stanfording:
> I can do this if there is a real need from the keras side.

Yes, please! Because scipy is important to Keras and the machine learning community.

eli-osherovich:
> > I can do this if there is a real need from the keras side.
> 
> Yes, please! Because scipy is important to Keras and the machine learning community.

[...]

PR #334, opened 2021-01-21
Removed SciPy dependency.

eli-osherovich:
### PR Overview

This is the second (last) part of the effort to remove SciPy dependency (#322)

Note: currently two tests are failing because of a bug tensorflow/addons#2357. Since the issue is not critical -- it only affects boundaries, we can adjust unit tests to follow the buggy behavior of `tensorflow-addons`.

### DIFF
--- a/keras_preprocessing/image/affine_transformations.py
+++ b/keras_preprocessing/image/affine_transformations.py
@@ -5,13 +5,6 @@ import numpy as np
 from .utils import array_to_img, img_to_array
 
 try:
-    import scipy
-    # scipy.ndimage cannot be accessed until explicitly imported
-    from scipy import ndimage
-except ImportError:
-    scipy = None
-
-try:
     from PIL import Image as pil_image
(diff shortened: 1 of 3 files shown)


eli-osherovich:
I think, there's no reason to keep "external" dependencies. I would suggest
to get rid of numpy too.

eli-osherovich:
In fact we do not need anything but TF here. Instead of `tensorflow_addons` one can use here `tf.raw_ops.ImageProjectiveTransformV3`

Dref360:
Once tensorflow/tensorflow#46637 is merged I think we can merge this pr.

Thank you!

eli-osherovich:
> Once [tensorflow/tensorflow#46637](https://github.com/tensorflow/tensorflow/issues/46637) is merged I think we can merge this pr.
> 
> Thank you!

They will not approve it because it breaks internal tests (that are relying on the current behavior).
Should we update our unit tests to rely on the TF behavior?

PR #344, opened 2021-07-02
remove dtype from np.asarray

Emekaborisama:
The recent version of numpy automatically select a data type and accept one positional argument. The dtype return a dependencies error 

'in asarray return array(a, dtype, copy=False, order=order) TypeError: __array__() takes 1 positional argument but 2 were given -->'

### PR Overview


Emekaborisama:
@googlebot I signed it!

Dref360:
This is a bug in Pillow, they are already working on a fix.

https://github.com/python-pillow/Pillow/pull/5572

PR #352, opened 2022-02-07
More robust random seeding using a numpy generator

psomers3:
### Summary
Only initialize the seed once on iterator creation. Maybe an option to reset it could be added. 
**Update** This pull request actually needs more work, because it seems like if you want to synchronize two generators (i.e. images and masks) the geomtric tranformations in the ImageDataGenerator can very easily be non deterministic as well.

### Related Issues
 #351 - not entirely related to this, but in the same direction of truly reproducable results.
### PR Overview


psomers3:
I saw somewhere this repository is no longer active, so I will close this as it doesn't fully fix any problem.

